PR #274, opened 2023-03-23
Custom docker support

smahmed776:
**Chnages:**
+ Introduced a node js server called `Worker-manager` which is responsible for preparing the environment before the script run and cleanup after the run.
+ Developers can now define a custom docker container for their nodes to run their node on that container and also run the container with required arguments (if any).
+ Modified electron server to be aligned with the worker-manager server.
+ Ensured effectiveness of `.dockerignore` file as previously this file was not correctly implemented. As a result, our docker image size is reduced from 5gb to around 2.7gb
+ Modified `flojoy` script with some color modification and re-ordering of the commands.


jalium:
Is this issue now nulled by [this PR](https://github.com/flojoy-io/studio/pull/309)?

jackparmer:
> Is this issue now nulled by https://github.com/flojoy-io/studio/pull/309 to remove docker?

Indeed - I think we can simply close this one

PR #306, opened 2023-04-23
Refactor script

### LINKED ISSUE #304: [QE] flojoy bash script is practically unusable
The installation and first-time UX is still far from smooth.
Here are the things that I noticed:

- [x] This should be Mac/Unix. Hide the GitPod option because this never worked.
image

- [x] The exact command should always be there to copy/paste (eg, `cd studio`). There should never be any guesswork left to the user.
image

- [x] Installing `gum` failed fantastically with homebrew - should we drop this dependency if it's too hard to support?

- [x] `Credentials` is spelled wrong here ❗It should also tell the user how to resolve this.
image

- [x] All lines have `\e[0m` printed after them ❓ 
image

- [x] should give you the commands to test your Python and pip versions. Should also check for these version dependencies programmatically at the start of the script, and error if they are not met!!!! 
image

- [x] The terminal text color is not set back to the default:
image

- [x] There's no feedback while the python and node packages are being installed - ideally they should be a line-by-line for each failed or successful python/node package install. The npm install hung for me then, ate up 4GB of RAM, then crashed my computer - no feedback printed about what was happening.
image

> ❌ Failed to install Python packages, check if pip and Python not lower than v3.10 is installed in your local machine or Try running this script again. \e[0m

- [x] This error is not acceptable.  Why is Try capitalized? "if pip and Python not lower" is not proper English. Instead of not lower, you can use >=. The Python version is v3.10, but that does not mean that the pip version is 3.10 (they're on a different versioning system!). The syntax color is not set back to the default from red and there should be 2 newlines after this error. No details are printed, so it's impossible where the failure was.

> No virtual env provided. Will use virtual env from /Users/jackparmer/.flojoy/venv directory
- [x] Why do we need a default venv? Why force the user to use any venv if they don't specify one? A venv should not be required to run the software.

- [x] seems like startup and flojoy are essentially the same shell script, just one doesn't use GUM. need to keep things DRY 🏜️ 

- [x] This error appears as soon as I run the script. Bad UX/first impression:
> Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.

smahmed776:
Resolved #304 
**Changes:
- Removed `Gum` from the startup script.
- Added a new script to check if all pre-required dependencies are installed like Python, Node.js, etc.
- Removed installation guidelines from README.md and added a reference link https://docs.flojoy.io/getting-started/install/ for installation guidelines.


jackparmer:
@smahmed776 what can't the shell script start Redis itself, if it doesn't detect a Redis process?

image

jackparmer:
still getting this error in multiple locations:

> zsh: parse error near `&&'

image

image

image

image

dev-nested:
I pushed a fix to the `flojoy` starter shell script. But for this to work redis has to be running again.

I had to remove the code that checked if Redis has started. 
For some reason, the shell is running the next commands before executing anything of ttab.

**A suggestion to improve how it handles Redis:**
- Check if redis-cli is there, otherwise quit installation saying the user needs to install redis. In fact the script can do that for node and python as well at the very beginning.
- Check if redis is already running
- If redis is not running, run it
  - The code that checks whether redis has started or not, is not working..
  - The installer has to wait for redis to start before it can run the rq workers, so we have to find another way to do that

I don't think we need to shutdown an already running redis. Still if want to keep that support, we can use the cli argument `-r`, which will mean that, if redis is running then the installer should stop it and run redis again.

mdasifhasan (approved):
I tested on my local mac, its working fine

PR #309, opened 2023-04-29
Remove docker

smahmed776:
Changes: 
- Removed Docker from CI.
- Using `pm2` package to run backend services.
- Executing startup script to install packages and run necessarily required operations.


jackparmer:
please separate removing docker cruft from other in-flight startup script optimizations (like introducing `pm2`) 
PRs should only do 1 thing in scope to make them easy to review and roll back if needed - this PR should only be removing legacy docker stuff

smahmed776:
> 

PM2 is used only in CI here not in the startup script to replace Docker and run all the services. I think this is inseparable. If we just remove Docker then the CI test will not pass by any means.

PR #313, opened 2023-05-01
Format Python and TypeScript for open-source

itsjoeoui:
Since we are open-source, we should have a code formatting standard for both Python and TypeScript, which will be beneficial for other contributors.


izi-on (approved):
Nice

JeffDotPng (approved):
:+1:

PR #334, opened 2023-05-04
Fix saving on non-chrome browsers and disable Save As when not supported

JeffDotPng:
Currently opening a save file picker only works on Chrome >= 86, so on non chrome based browsers the button will just be disabled.


jackparmer:
💃 after test fix

PR #346, opened 2023-05-04
Update index.html file

smahmed776:
Removed unused index.html file from /public/ directory and updated index.html file in the root directory


itsjoeoui (approved):
LGTM

PR #362, opened 2023-05-10
Update BAR node

smahmed776:
Fixed a bug in BAR node.


smahmed776:
Merging this as its corresponding one in the nodes repo is merged already!

PR #366, opened 2023-05-11
Fix local test

izi-on:
Backend now connected for tests


package.json
```diff
@@ -46,11 +46,11 @@
     "backend": "python manage.py runserver",
     "rq-watch": "OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES rq worker flojoy-watch",
     "rq-flojoy": "cd PYTHON && OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES rq worker flojoy",
-    "start-project": "concurrently \"npm run start\" \"npm run backend\" \"npm run rq-watch\" \"npm run rq-flojoy\"",
+    "start-project": "CI=true concurrently \"npm run start\" \"npm run backend\" \"npm run rq-watch\" \"npm run rq-flojoy\"",
```
smahmed776, line 49: Instead, we can create a new script for running in CI mode. As developers would like to see their original function running in local mode instead of the Mock version.

smahmed776 (changes requested):
Overall looks good, however, requested a change in script...

smahmed776 (approved):

PR #369, opened 2023-05-12
add -t option for running CI=true (npm run start-project:ci) for runn…

izi-on:
title says it all

### DIFF
--- a/flojoy
+++ b/flojoy
@@ -164,6 +164,10 @@ do
         is_ci=true
         shift
         ;;
+        -t)
+        test_ci=true #run with CI=true, allows to properly run tests locally 
+        shift
+        ;;
         *) # unknown option
         echo "Unknown option: $1"
         helpFunction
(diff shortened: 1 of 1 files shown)


flojoy
```diff
@@ -352,6 +356,11 @@ sleep 1
 # start project services
 if [ -n "$venvCmd" ]; then
    $venvCmd && npm run start-project
+
+elif $test_ci; then
```
smahmed776, line 360: This can be written as follows:
```
if $test_ci; then
     start_project_cmd=npm run start-project:ci
else 
      start_project_cmd=npm run start-project
fi
 # start project services
if [ -n "$venvCmd" ]; then
   $venvCmd && $start_project_cmd
else
    $start_project_cmd
fi
```

smahmed776: $test_ci variable check should be checked independently.

smahmed776 (changes requested):

smahmed776 (approved):
LGTM!

PR #370, opened 2023-05-12
New nodes: R datasets, Table [Plotly]

smahmed776:
Two new nodes were added: R Dataset and Table


jackparmer:
Please always include a handful of app and ctrl panel screenshots!

In the ctrl panel, please make sure that the table uses colors from the Figma flojoy brand guide

plotly's table looks pretty bad in small (thumbnail) formats - in the flow chart app, probably want to show only preview the first 2 columns and 4 rows for a table node

jackparmer:
Please also move the example app over to the docs repo https://docs.flojoy.io/

We're going to keep example apps in the docs repo instead, and delete the apps repo

PR #377, opened 2023-05-13
364 refactor sidebar

izi-on:
refactored the old sidebar to use the new common sidebar. 

important notes:

- the tests had to be adjusted: for 04 and 05, you need to filter by visible, since we are using the same component template for the sidebar in the script and ctrl tab (they aren't unmounted when navigating tabs)
- the new exported COMMAND manifest is reduced to an object with key value pairs which makes it hashable, now we don't scan the entire array anymore for a match. (Note: the json is left unchanged, just using .reduce on the array to convert it to an object with fields)


itsjoeoui (approved):
lgtm!

PR #382, opened 2023-05-14
Wrap node names

Jin1L:


itsjoeoui:
@Jin1L Looks like you need to regenerate the snapshot for the CI to pass. 

`npm test -- -u` should do the job

PR #388, opened 2023-05-15
fixed a bug related to cursor

Orisu179:
When hoving over the pencil and close icons, the cursor will now changes to "point" from "grab"


itsjoeoui (approved):
lgtm!

PR #390, opened 2023-05-15
Set api key

Jin1L:


smahmed776 (changes requested):

src/feature/flow_chart_panel/views/API_keyModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+    },
+  };
+
+  const handleApiKeyChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setApiKey(e.target.value);
+  };
+
+  const sendApiKeyToDjango = async (apiKey: string) => {
+    try {
+      const response = await fetch("http://localhost:8000/api/set-api", {
```
smahmed776, line 93: We should use a variable for the backend server URL instead of hard-coding it. For reference: https://github.com/flojoy-io/studio/blob/main/src/services/FlowChartServices.ts#L9

src/feature/flow_chart_panel/views/API_keyModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+      console.error("An error occurred:", error);
+    }
+  };
+
+  const handelSendApi = () => {
+    sendApiKeyToDjango(apiKey);
+  };
+
+  return (
+    <ReactModal
```
smahmed776, line 117: I think it's better to use `@mantine` for UI components as we're using it everywhere in the project and remove 'react-modal'.

src/feature/flow_chart_panel/views/API_keyModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        console.error("Request failed:", response.status);
+      }
+    } catch (error) {
+      console.error("An error occurred:", error);
+    }
+  };
+
+  const handelSendApi = () => {
+    sendApiKeyToDjango(apiKey);
```
smahmed776, line 113: We should handle if `apiKey` is empty or not here.

src/feature/flow_chart_panel/views/API_keyModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+            width: 23,
+          }}
+        />
+      </button>
+
+      <div className={classes.container}>
+        <div className={classes.title}>
+          <div>API Key</div>
+          <input type="text" onChange={handleApiKeyChange} value={apiKey} />
+          <button onClick={handelSendApi}>Submit</button>
```
smahmed776, line 138: We may keep the submit button disabled until some values are given to input.

smahmed776 (changes requested):
Looks good! I have left some comments, resolving them would be appreciated!

src/feature/flow_chart_panel/views/API_keyModal.tsx
youngsun4786: We have referenced the structure in other files such as NodeModal.tsx and KeyboardShortCutModal.tsx and they were utilizing the ReactModal which adapted the styles function from `@mantine/core`. Should we just go with regular Modal imported from `@mantine` then?

youngsun4786: @smahmed776 So should we update them to Mantine modal for other files as well and as you mentioned resolve them in a separate PR?

smahmed776: @youngsun4786 yes please, if you can. You may create a new issue and work on that...

youngsun4786: @smahmed776 Sounds good!

smahmed776 (approved):
LGTM!

PR #391, opened 2023-05-16
Develop

jackparmer:


requirements.txt
```diff
@@ -21,3 +21,4 @@ requests==2.28.1
 # Testing
 pytest==7.1.3
 black==23.3.0
+ticlib
```
itsjoeoui, line 24: We should probably pin this as well!

src/feature/common/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,259 @@
...
+      );
+    }
+
+    var content;
+
+    if (textInput !== "") {
+      //case 1: name is included in the string of the section node or leaf node
+      if (node.title.toLowerCase().includes(textInput.toLocaleLowerCase())) {
+        //case 1.1: node has children (is a section)
+        if (node["child"] != null && !("key" in node)) {
```
itsjoeoui, line 117: Looks like this should be `!==` instead? @izi-on

src/feature/common/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,259 @@
...
+  sections: Node;
+  leafNodeClickHandler: leafClickHandler;
+  manifestMap: any; //Key value pair object
+}) => {
+  const [textInput, handleChangeInput] = useState("");
+  const addButtonClass = useAddButtonStyle();
+  const { classes } = useSidebarStyles();
+
+  //this function will create the sections to be rendered according to the search input
+  const renderSection = (textInput: string, node: Node, depth: number) => {
```
itsjoeoui, line 102: @izi-on Can you also add the return type of this function please

src/feature/common/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,259 @@
...
+              />
+            );
+          }
+        }
+      }
+
+      //case 3: no search input
+    } else {
+      //case 3.1: node has children (is a section)
+      if (node["child"] != null && !("key" in node)) {
```
itsjoeoui, line 171: Should be `!==`, there are a couple `==` below that should be changed as well

src/feature/common/defaultPlotLayout.ts
```diff
@@ -1,7 +1,7 @@
 import { Layout } from "plotly.js";
 
 const styledPlotLayout = (theme: "light" | "dark") => {
-  const plotFeatureColor = "#282c34";
+  const plotFeatureColor = theme === "light" ? "#282c34" : "#fff";
```
itsjoeoui, line 4: We may want to use the theme we defined instead of hard-coding

itsjoeoui (changes requested):

itsjoeoui (dismissed):

requirements.txt
```diff
@@ -13,12 +13,11 @@ PyYAML==6.0
...
+uptime==3.0.1
+schedule==1.2.0
 djangorestframework==3.14.0
 python-dotenv==0.21.0
 requests==2.28.1
-# Testing
 pytest==7.1.3
 black==23.3.0
-ticlib
+ticlib==0.2.2
```
smahmed776, line 23: This is a node-specific dependency and shouldn't be here.

smahmed776 (changes requested):

smahmed776 (approved):

PR #397, opened 2023-05-16
Correct steppernode

Angchau:
@smahmed776 I updated the COMMAND MANIFEST file in order to include my new nodes STEPPER DRIVER TIC and KNOB


smahmed776 (approved):
Look good.

PR #407, opened 2023-05-17
Added subtle prompt to rerun app

Jin1L:


smahmed776 (changes requested):
Good approach, I've left a couple of comments.

smahmed776 (approved):

Issue #395, opened 2023-05-16
Allow for easier time with parameter importing


Orisu179:
Currently, if a user wants to import int/float from the control panel, they would need to convert them manually from string. For example:

```Python
from flojoy import flojoy, DataContainer

@flojoy
def add(dc_inputs, params):
     value1 = float(params["val1"])
     value2 = float(params["val2"])
     sum = value1 + value2
     return dc_inputs

```

It would be nice if the user doesn't have to convert them and can use them right away.

PR #423, opened 2023-05-18
Added Array input to the controls panel

Orisu179:
Currently it works with the develop branch. A problem is that the user can still enter string values in the input box. Other than that it should work as expected. **This PR doesn't support a type system yet**, it's just laying the groundwork for now.

Some notes:
- Introduced types to the node props
- The types will be sent to the backend where the decorator will convert the data from string into the specified type
- The details of the above will be fixed and update in the future in #395 (which is what I'll be working on next)
- This PR is made along with a PR to the python repo in the backend: https://github.com/flojoy-io/python/pull/25


itsjoeoui (changes requested):

package.json
```diff
@@ -10,6 +10,7 @@
     "@tabler/icons-react": "^2.17.0",
     "@tisoap/react-flow-smart-edge": "^3.0.0",
     "cors": "^2.8.5",
+    "eslint-config-react-app": "^7.0.1",
```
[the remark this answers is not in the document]
itsjoeoui: Seems like this is made for create-react-app but we are no longer using it. Can you remove it please?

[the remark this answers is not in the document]
Orisu179: ok

Orisu179:
I realized from the test that I would need to change more parts of the code, so this isn't ready yet

src/feature/controls_panel/ControlsTabView.tsx
```diff
@@ -101,6 +105,7 @@ const ControlsTab = ({ results }: ControlsTabProps) => {
         functionName: (ctrl.param! as CtrlManifestParam).functionName,
         param: (ctrl.param! as CtrlManifestParam).param,
         value: val,
+        ValType: tp,
```
smahmed776, line 108: `camelCase` is preferable here.

src/feature/controls_panel/manifest/CONTROLS_MANIFEST.ts
```diff
@@ -57,6 +58,11 @@ export const CTRL_TREE = {
               key: "NUMERIC_INPUT",
               child: null,
             },
+            {
+              title: "Array Numeric Input",
```
smahmed776, line 62: Should it be `Numeric` or `Numberic`? Consistency should be maintained throughout all the files.

src/feature/controls_panel/views/control-component/ControlComponent.tsx
```diff
@@ -361,12 +375,33 @@ const ControlComponent = ({
...
+              );
             }}
             value={currentInputValue || numberInput || 0}
           />
         </Box>
       )}
+      {ctrlObj.name === ControlNames.ArrayNumbericInput && (
+        <Box className={classes.inputBody}>
+          <input
+            type="string"
```
smahmed776, line 391: Should be `text` for `string` values IMO

src/feature/common/types/ParamValueType.ts
```diff
@@ -0,0 +1,10 @@
+export enum ParamValueType {
```
itsjoeoui, line 1: Generally we prefer string union over Enum in TS

itsjoeoui (changes requested):

src/feature/flow_chart_panel/components/node-edit-menu/ParamField.tsx
```diff
@@ -23,13 +22,14 @@ const ParamField = ({
   const { updateCtrlInputDataForNode } = useFlowChartState();
   const handleChange = (value: string) => {
     updateCtrlInputDataForNode(nodeId, paramId, {
-      functionName,
+      functionName: functionName,
```
smahmed776, line 25: We can use just `functionName` here as the key and value both are the same.

src/feature/flow_chart_panel/components/node-edit-menu/ParamField.tsx
```diff
@@ -23,13 +22,14 @@ const ParamField = ({
...
       param: paramId,
-      value,
+      value: value,
```
smahmed776, line 27: Same here, for ES6 we don't need to set value explicitly if the key and value are the same.

src/feature/common/types/ParamValueType.ts
Orisu179: Got it

smahmed776 (changes requested):
Great job, just some cosmetic things ...

src/feature/controls_panel/manifest/CONTROLS_MANIFEST.ts
Orisu179: that's a typo, thanks

src/feature/controls_panel/ControlsTabView.tsx
```diff
@@ -86,7 +86,11 @@ const ControlsTab = ({ results }: ControlsTabProps) => {
     // }
   };
 
-  const updateCtrlValue = (val: string, ctrl: CtlManifestType) => {
+  const updateCtrlValue = (
+    val: string,
+    ctrl: CtlManifestType,
+    camelCase: ParamValueType
```
smahmed776, line 92: I apologize that my previous comment was not clear to you, I meant to change `ValType` to match the format `camelCase` as follows: `valType`

Orisu179: Sorry, my bad for not asking for a clarification, I didn't know what camelCase meant and just assumed it was a rename.

src/feature/controls_panel/views/control-component/ControlComponent.tsx
Orisu179: This is doable but the types in manifests-latest.json are "string" instead of "text". Should I change them as well?

src/feature/controls_panel/ControlsTabView.tsx
smahmed776: No problem at all, `camelCase` is indeed a naming convention commonly used in programming, [This article](https://dev.to/prahladyeri/underscores-camelcasing-and-pascalcasing-the-three-naming-conventions-every-programmer-should-be-aware-of-3aed) could be helpful for you to understand naming conventions used in programming.

smahmed776 (approved):
Lgtm

src/feature/controls_panel/ControlsTabView.tsx
```diff
@@ -18,8 +18,8 @@ import { useControlsTabEffects } from "./ControlsTabEffects";
 import { CtrlOptionValue } from "./types/ControlOptions";
 import { ResultsType } from "@src/feature/results_panel/types/ResultsType";
 import { createStyles } from "@mantine/styles";
-import { useMantineTheme } from "@mantine/styles";
-
+import { useMantineColorScheme, useMantineTheme } from "@mantine/styles";
```
JeffDotPng, line 21: Unused imports, can be removed

src/feature/flow_chart_panel/hooks/useAddNewNode.tsx
```diff
@@ -60,6 +61,7 @@ export const useAddNewNode = () => {
                 funcName === "CONSTANT"
                   ? nodeLabel
                   : params![param].default?.toString(),
+              valType: "unknown" as ParamValueType,
```
JeffDotPng, line 64: Could you not get the correct type from `params`? e.g `params[param].type`

Orisu179: RIght! I wrote this as a placeholder and forgot about it. Tysm!

itsjoeoui (approved):

PR #425, opened 2023-05-18
Reduce Plotly.js bundle size

mehedihasansabbir220:
In this commit, I optimized the Plotly.js bundle size by creating a custom bundle. The full package of Plotly.js was previously shipped, resulting in a larger file size.

To create the custom bundle, I followed these steps:
- Installed Plotly.js and its dependencies.
- Ran the custom-bundle command with the necessary options to include only the required trace types and exclude transforms.
- Specified an output filename for the custom bundle.

Additionally, I updated the import statements in the code to include specific modules:
- Imported Data and Layout from "plotly.js".
- Imported Plot, PlotParams from "react-plotly.js".
- Imported PlotData, PlotType from "plotly.js".

These changes will help reduce the bundle size and optimize the application's performance.

Please review and merge this commit.


jackparmer:
nice!
can you please add a screenshot for each of the viz node types, to make sure that they are still working?

mehedihasansabbir220:
image
@jackparmer

smahmed776:
Please add screenshots for all of the visual nodes to ensure the introduced changes not breaking any of the existing visual nodes.

jackparmer:
@mehedihasansabbir220 do you plan on finishing this or can we close it? thanks!

itsjoeoui:
> @mehedihasansabbir220 do you plan on finishing this or can we close it? thanks!

Jeff tested it, and it does not actually reduce the bundle size, the approach here is not the one mentioned in the docs.

I suggest we re-work this based on the latest develop branch.

PR #433, opened 2023-05-19
Updated expand button and test resolved

Jin1L:
- Added expand icon and updated css
- Updated expand button and test resolved

### DIFF
--- a/src/feature/flow_chart_panel/views/NodeExpandMenu.tsx
+++ b/src/feature/flow_chart_panel/views/NodeExpandMenu.tsx
@@ -30,12 +30,8 @@ export const NodeExpandMenu = ({
 }: NodeExpandMenuProps) => {
   const { isExpandMode, setIsExpandMode } = useFlowChartState();
 
-  const canEditNode = clickedElement
-    ? Object.keys(clickedElement.data.ctrls).length > 0
-    : false;
-
   const onSelectionChange = () => {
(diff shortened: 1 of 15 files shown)


src/feature/flow_chart_panel/views/NodeExpandMenu.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (!clickedElement || !canEditNode) {
+      setIsExpandMode(false);
+    }
+  };
+
+  useOnSelectionChange({ onChange: onSelectionChange });
+
+  return (
+    <Box pos="relative" data-testid="node-modal">
+      {clickedElement && canEditNode && isExpandMode && (
```
JeffDotPng, line 47: This makes it so you can't inspect nodes that aren't editable, should remove `canEditNode` from this condition

itsjoeoui (changes requested):
Seems like the nodes that do not have expand functionality also have the expand icon. Can yo u only display that icon on nodes that are expandable?

src/feature/flow_chart_panel/views/NodeExpandMenu.tsx
itsjoeoui: Aha yes this exactly.

itsjoeoui (approved):

PR #436, opened 2023-05-23
update delete node keystroke

scorinaldi:
For the record: I am working on a ThinkPad IBM / Ubuntu / and running the studio in my browser. 

When I try to use Ctrl-D to delete a node as directed in the menu, it doesn't work. Backspace does work though. This change reflects that.


itsjoeoui (approved):
lgtm! I am currently re-working all keyboard shortcuts, will merge this for now but I will make an update in the near future.

PR #439, opened 2023-05-23
replaced python codes to github links

youngsun4786:


src/feature/flow_chart_panel/views/NodeModal.tsx
```diff
@@ -17,9 +19,36 @@ const NodeModal = ({
...
+    });
+  });
+
+  let LINK = "";
+  switch (nodeDataLabel) {
+    case "END":
+      LINK = `${GLINK}/${nodeCategory}/${nodeTypeLink}/${pythonString}`;
+      break;
+    case "OBJECT_DETECTION":
+      LINK = `${GLINK}/AI_ML/${nodeDataLabel}/${pythonString}`;
```
itsjoeoui, line 46: We should probably not hardcode "AL_ML" here I think.

youngsun4786: Oh okay! The file name was acquired using the title property from COMMANDS_MANIFEST.ts file and the title name was something like "AI and Machine Learning". What alternative do you suggest in this situation?
Screenshot 2023-05-23 at 2 59 16 PM

youngsun4786: I agree with hard coding being a malpractice here.

itsjoeoui: @smahmed776 Shall we just change the title to stay consistent in this case?

Jin1L: @itsjoeoui Similar thing for CLOUD_DATABASE

smahmed776: @itsjoeoui  We can add a key field for each parent category and update all folders not matching the key(if any) and keep the title as it is because it looks better with spacing and full name in the sidebar than these shorthands IMO. Also, we need to ensure each node folder follows its file name like `LINSPACE/LINSPACE.py`. This should resolve all the issues.

note: We need to update docs in the docs repo to force app builders to follow this convention.

itsjoeoui: Sounds good! I think we can start with that such that we can have a better structure, I can take a look in a bit

youngsun4786: Sounds good! I'll be working on it today!

youngsun4786:
> 



> 

@smahmed776 Advised me to start working on the UX for the nodes along with this PR, hence the progress so far looks like this if you were referring to the UX:
Screenshot 2023-05-24 at 5 00 04 PM

The links do work except I still have to change the structure for manifest a bit and rename some files.

jackparmer:
can you add some screenshots please @youngsun4786 ?

youngsun4786:
@jackparmer Hey Jack! I apologize for a late response. 
I've noticed that on there was a secondary button color on https://www.flojoy.io/brand so I made the corresponding changes.
- modified the logic for getting file paths for the github link.
- added the colors to code examples & incorporated `JSONTree` for jsons code.

https://github.com/flojoy-io/studio/assets/83793112/b31de4d3-51a7-49c4-9ec3-c5ae605ae821

src/feature/flow_chart_panel/FlowChartTabView.tsx
```diff
@@ -205,16 +207,18 @@ const FlowChartTab = () => {
...
+    let nodeFileName =
+      selectedNode?.data.label === defaultPythonFnLabel ||
+      selectedNode?.data.type === defaultPythonFnType
         ? "..."
-        : PYTHON_FUNCTIONS[selectedNode?.data.label + ".py"];
+        : selectedNode?.data.label + ".py";
 
     if (selectedNode.data.func === "CONSTANT") {
-      pythonString = PYTHON_FUNCTIONS[selectedNode.data.func + ".py"];
+      nodeFileName = selectedNode.data.func + ".py";
```
smahmed776, line 217: This line applies for all the nodes not only `CONSTANT`

src/feature/flow_chart_panel/views/NodeModal.tsx
```diff
@@ -18,30 +189,111 @@ const NodeModal = ({
...
+    const path: string[] = getPath(CMND_TREE, nodeType)!;
+    let pathLength = path.length;
+    if (nodeType === "STEPPER") pathLength = -1;
+    nodeCategory = path[0];
+    for (let i = 0; i < pathLength; i++) {
+      LINK += `/${path[i]}`;
+    }
+  }
+
+  switch (nodeCategory) {
```
smahmed776, line 217: Can we set a name convention for all categories/sub-categories/files so that we can get rid of this `switch` method and we don't have to change here again in case of new category/file is introduced?

src/feature/flow_chart_panel/views/NodeModal.tsx
```diff
@@ -61,22 +313,27 @@ const NodeModal = ({
...
-      <h3>Python code</h3>
+      <div style={{ display: "flex" }}>
+        <h2 style={{ fontSize: 22, marginTop: 2, marginBottom: 0 }}>
+          Python code
+        </h2>
+      </div>
       <SyntaxHighlighter
         language="python"
-        style={colorScheme === "dark" ? srcery : docco}
+        style={colorScheme === "dark" ? darkFlojoy : lightFlojoy}
```
smahmed776, line 323: This line gives a type error to me

smahmed776 (changes requested):

smahmed776 (approved):

PR #440, opened 2023-05-23
Fixed some bugs relating to detecting parameter types

Orisu179:
- now when importing values from the control panel, the user doesn't have to convert them from string
- Added type values in the RECIPES.ts file
- This PR is made with https://github.com/flojoy-io/python/pull/25, which should be merged first before this


smahmed776:
Add this `flojoy==0.1.3.dev1` to requirements.txt and update the PR please.

Orisu179:
> Add this `flojoy==0.1.3.dev1` to requirements.txt and update the PR please.

Done

PR #441, opened 2023-05-23
305 adding controls for default queue timeout and default deadtime between node execution

izi-on:


src/feature/controls_panel/ControlsTabView.tsx
```diff
@@ -91,7 +67,7 @@ const ControlsTab = ({ results }: ControlsTabProps) => {
     tp: ParamValueType
   ) => {
     const manClone = clone(ctrlsManifest);
-    manClone.forEach((c, i) => {
+    manClone.forEach(() => {
```
smahmed776, line 70: This function is doing nothing, we can get rid of it.

src/feature/flow_chart_panel/views/SettingsModal.tsx
```diff
@@ -0,0 +1,120 @@
+import ModalCloseSvg from "@src/utils/ModalCloseSvg";
```
smahmed776, line 1: Do we have a UT for this component? we should write UT's side by side for newly introduced components.

src/hooks/useSettings.ts
```diff
@@ -0,0 +1,50 @@
+import { atom, useAtom } from "jotai";
```
smahmed776, line 1: Do we have a UT for this hook?

src/hooks/useSettings.ts
```diff
@@ -0,0 +1,50 @@
...
+import { useEffect } from "react";
+
+export type Settings = {
+  title: string;
+  key: string;
+  type: string;
+  group: string;
+  value: any;
```
smahmed776, line 9: We should avoid using `any` type unless there is a special need for it.

src/hooks/useSettings.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const useSettings = () => {
+  const [nodeDelay, setNodeDelay] = useAtom(nodeDelayDefault);
+  const [maximumRuntime, setMaximumRuntime] = useAtom(maximumRuntimeDefault);
+  const [settingsList, setSettingsList] = useAtom(settingsListDefault);
+  useEffect(() => {
+    console.log("here");
+    setSettingsList([
+      {
+        title: "Node Delay (seconds)",
```
smahmed776, line 25: We can move these constant values to `settingListDefault` variable and only update the necessary fields here.

src/hooks/useSettings.ts
```diff
@@ -0,0 +1,50 @@
...
+  key: string;
+  type: string;
+  group: string;
+  value: any;
+  setValue: any;
+};
+
+const nodeDelayDefault = atom(0);
+const maximumRuntimeDefault = atom(3000);
+const settingsListDefault = atom([{} as Settings]);
```
smahmed776, line 15: Instead this could be written as follows: `const settingsListDefault = atom<Settings[]>([]);

smahmed776 (changes requested):

src/hooks/useSettings.ts
smahmed776, line 1: Do we have a UT for this component? we should write UT's side by side for newly introduced components.

smahmed776 (approved):
Lgtm

PR #448, opened 2023-05-24
Setup Sentry

itsjoeoui:
Per title, also nuked all the manifest.json since they are auto generated anyways.


smahmed776 (approved):
Looks good, I recommend adding a description for `-S` parameter inside `helpFunction()` in flojoy script.

PR #459, opened 2023-05-25
Reduce highlightjs and plotly.js bundle size

JeffDotPng:
Changes:
- Bumped `plotly.js` version to 2.23.2
- Added a command in package.json to generate a custom bundle for `plotly.js` (1.1mb gzipped -> 584kb)
You can choose to use the custom bundle or the full size one in `PlotlyComponent.tsx` by commenting out a line. (For dev purposes, the full size bundle is fine, when building for production we can use the custom one)

- Reduced `highlight.js` bundle size by only importing languages that are needed (288kb gzipped -> 10kb)


itsjoeoui:
For dev, can we also stick to our custom bundle as well? We just add whatever plot we want as we need more, stick with our custom bundle

JeffDotPng:
> For dev, can we also stick to our custom bundle as well? We just add whatever plot we want as we need more, stick with our custom bundle

You can, we can just make it the default, you'll just have to regenerate it every time you nuke `node_modules` (which hopefully isn't that often?)

itsjoeoui:
@smahmed776 Can you try this on your machine and see if it works right out of the box? Otherwise it looks good to me

smahmed776:
> @smahmed776 Can you try this on your machine and see if it works right out of the box? Otherwise it looks good to me

Will do

itsjoeoui:
> > @smahmed776 Can you try this on your machine and see if it works right out of the box? Otherwise it looks good to me
> 
> Will do

Thanks!

itsjoeoui (approved):
lgtm!

PR #466, opened 2023-05-26
fix a bug with node reference

Orisu179:
- Now it works as expected:
image


src/feature/flow_chart_panel/components/node-edit-menu/ParamField.tsx
```diff
@@ -60,6 +60,13 @@ const ParamField = ({
       );
     case "select":
       return <Select onChange={handleChange} data={options!} value={value} />;
+    case "node_reference":
```
smahmed776, line 63: Node reference shouldn't be a simple text input rather it should be a dropdown of ID's of all nodes in the flowchart IMO

smahmed776 (changes requested):

src/feature/flow_chart_panel/components/node-edit-menu/ParamField.tsx
Orisu179: Agreed, but it would take sometime for me to figure that out. Right now any node with references doesn't work at all so this is more like a placeholder. I plan to work on it after I finish my current task

src/hooks/useFlowChartGraph.ts
Orisu179, line 1: Used for loop instead of filter so it only iterates through once, Alternatively there can be two filters, but due to updating every time a node is dragged, I figured this would be better

smahmed776: Loop is better, however it can be wrapped with `useMemo` as follows:
```javascript
 const { selectedNodes, unSelectedNodes } = useMemo(() => {
    const selectedNodes: Node<ElementsData>[] = [];
    const unSelectedNodes: Node<ElementsData>[] = [];
    for (const n of nodes) {
      if (n.selected) {
        selectedNodes.push(n);
      } else unSelectedNodes.push(n);
    }
    return { selectedNodes, unSelectedNodes };
  }, [nodes]);
  ```

File: src/feature/flow_chart_panel/components/node-edit-menu/ParamField.tsx [hunk 2]
```diff
@@ -60,6 +62,10 @@ const ParamField = ({
...
+      return (
+        <Select onChange={handleChange} data={otherNodeIds!} value={value} />
```
smahmed776, line 67: We can use nullish coalescing `??` here to avoid eslint error. e.g `otherNodeIds ?? []`

smahmed776, line 67: Just showing ID's can be confusing as all node id's start with node function name so for duplicate nodes it'll show same name. We can display node labels for better UX.
e.g:
Screenshot (40)

smahmed776 (changes requested):

PR #469, opened 2023-05-26
Fix issue where rfInstance was not updated when running program

JeffDotPng:
Fixes the issue where the nodes/edges sent to the backend are not updated when running the program.


itsjoeoui (approved):
lgtm

PR #481, opened 2023-05-29
Mixpanel Telemetry

itsjoeoui:
@Orisu179 Next time please create a PR as you are working on it :D


Orisu179:
What this PR is missing: 

- Tracking program saved to Frontier
- Tests
- telemetry: false setting to the disk so that -T doesn't have to be called everytime the user runs the script

jackparmer (approved):

PR #486, opened 2023-05-29
Add .tgz for custom bundle

JeffDotPng:


flojoy
smahmed776, line 1: I think this line : https://github.com/flojoy-io/studio/pull/486/files#diff-5667e11402ebd91996d48e0b381a8a1d964cc3bc83135665e9b376214ab4afb9R179 should also be removed

JeffDotPng: Oops you're right, missed those, fixed now.

smahmed776 (approved):
LGTM!

PR #489, opened 2023-05-30
[DO NOT MERGE] - Implement event driven scheduling

dev-nested:
Currently we can see a visible lag when a node finishes and the next node starts. This PR tries to improve the time it takes to switch execution between nodes.

Currently the job scheduler (`Watch.py`) uses polling (every 100ms, configurable in frontend) to know whether a scheduled job has finished. In addition to this, the scheduler schedules jobs in batches. If it had scheduled 3 jobs in the previous batch, it'll first wait for all 3 jobs to finish. Then it'll find out which jobs have become newly ready and schedule them in a batch. 

This PR removes the polling and implements an event driven mechanism of scheduling. 

**Details**
The scheduler rq worker is removed. Now, Django maintains the topology in its memory and provides HTTP API to interact with it. Optionally the state can be saved in memory as well.

Tested with the default example app.

### Additional notes
- Has performance improved?
  - No. Its actually slower. 
- Why is it slower?
  - The main bottleneck is how fast Django processes the HTTP requests. In my local machine (mac) Django takes around (200ms ~ 300ms) to process the HTTP requests that are sent by the rq workers when they finish executing nodes. Compared to 100ms of polling period this is few times slower. So we actually see more lag between nodes with this implementation.
- Then what do we do with this PR?
  - I recommend to keep it open and unmerged for now as it has many improvements. We have planned to implement a different approach to scheduling where regular python functions will be executed in the scheduler itself (or one dedicated rq worker). This will achieve minimum lag between nodes because it'll then essentially be same as calling two functions in same program. There will be an option to mark a node needing independent rq worker which might be required for long running nodes. I recommend to keep this PR open because this might be a better starting point to implement the new approach.


dev-nested:
**Update:**
- The latest from today removes the need to have Redis as state manager, so with this pr Redis is now optional and disabled by default.
- Removed rq-watch scheduler. Now Django server itself enqueues jobs.
- So this is almost as fast as it gets on event driven mechanism where we use completely separate processes to execute each node.

**Performance Bottleneck:**
After investigating the timing with this new version, I found Django takes about 200ms-600ms to handle the API requests that the @flojoy. decorator calls. Whereas current polling based mechanism checks every 100ms by default (configurable in frontend). This makes this Django based event driven mechanism slower than the current.

**Next:**
The next gen of scheduling is to execute all regular python function nodes in the scheduler itself (Django/Fastapi/etc). It'll still use rq workers but only for nodes that need special rq workers or just a separate rq worker to run for long time.

PR #493, opened 2023-05-30
Runtime schema check for manifest-latest.json to ensure safety

itsjoeoui:
- Moved all the loading logic to `ManifestLoader.ts`
- Fix some types in nodes' manifest
- The app will now display an error page if the `manifest-latest.json` is bad, since Zod now does runtime type checking for the entire manifest file.
- Refactor the data loading using react-router's loader. This makes it more flexible in the future if we ever want to load the manifest data from a remote server


src/utils/ManifestLoader.ts
```diff
@@ -1,19 +1,71 @@
...
+
+export function getManifestParams() {
+  try {
+    const parsedManifest: Manifest = manifestSchema.parse(manifests);
+    return parsedManifest.parameters;
+  } catch (e) {
+    if (e instanceof ZodError) {
+      throw fromZodError(e);
+    } else {
+      throw new Error("something is seriously wrong");
```
JeffDotPng, line 49: Should probably throw a more descriptive error message (just in case this actually happens, then we would have a better idea of how to debug it)

smahmed776 (approved):

PR #494, opened 2023-05-30
462 refine all keyboard shortcuts

Jin1L:


src/Header.tsx
```diff
@@ -30,6 +32,23 @@ export type AppTab = "visual" | "panel" | "debug";
...
+
+  const debugShortCut = () => {
+    navigate("/debug");
+  };
+
+  const mainShortCut = () => {
+    navigate("/");
+  };
+
+  useKeyboardShortcut("shift", "C", () => ctrlsShortCut());
```
itsjoeoui, line 49: You can inline this by doing `() => {navigate("/debug")}`

youngsun4786: Done!

src/feature/flow_chart_panel/FlowChartKeyboardShortcuts.tsx
```diff
@@ -0,0 +1,94 @@
+import ReactFlow, {
```
JeffDotPng, line 1: Unnecessarily long file, no need to wrap the `ReactFlow` instance with this component. You could just keep `ReactFlow` component in `FlowChartTabView` as it was before, and just make a component that only calls hooks and returns an empty div, which avoids this long prop drilling and is significantly shorter.
```typescript
const FlowChartKeyboardShortcuts = () => {
  const { zoomIn, zoomOut, fitView } = useReactFlow();
  useKeyboardShortcut("ctrl", "=", zoomIn);
  useKeyboardShortcut("ctrl", "-", zoomOut);
  useKeyboardShortcut("ctrl", "1", fitView);

  useKeyboardShortcut("meta", "=", zoomIn);
  useKeyboardShortcut("meta", "-", zoomOut);
  useKeyboardShortcut("meta", "1", fitView);

  return <div />;
}
```

src/feature/flow_chart_panel/FlowChartTabState.ts
```diff
@@ -14,6 +14,7 @@ export function useFlowChartTabState() {
   const [pythonString, setPythonString] = useState("...");
   const [nodeFilePath, setNodeFilePath] = useState("...");
   const { setIsExpandMode } = useFlowChartState();
+  const [selectAllNodes, isSelectAllNodes] = useState(false);
```
JeffDotPng, line 17: This state seems pretty fragile (`FlowChartTabView.ts` line 199), what if the user selects all nodes using the shortcut, then clicks off to deselect them all? This state would still be true and the next time `ctrl + a` is pressed it wouldn't select all of the nodes. Instead of storing this as state, you could just compute it in the shortcut by checking if all of the nodes are already selected `nodes.every(n => n.selected)`

src/feature/flow_chart_panel/FlowChartTabView.tsx
```diff
@@ -191,6 +196,43 @@ const FlowChartTab = () => {
 
   const proOptions = { hideAttribution: true };
 
+  const selectAllNodesShortCut = () => {
```
JeffDotPng, line 199: Should be Shortcut, not ShortCut

src/feature/flow_chart_panel/views/ControlBar.tsx
```diff
@@ -215,6 +215,7 @@ const LoadButton = () => {
 
 const ControlBar = () => {
   const { states } = useSocket();
+  const { nodes, edges } = useFlowChartGraph();
```
JeffDotPng, line 218: Avoid calling this hook wherever possible, it makes the component rerender (with all of its children) every single time a node changes (so every time a node is dragged by a pixel in the editor). Only the buttons that need access to this state call this hook so the rerenders are isolated to them specifically, to minimize rerenders. You could just put the keyboard shortcuts in their respective buttons since they already have access to the state.

src/feature/flow_chart_panel/FlowChartTabView.tsx
```diff
@@ -229,61 +273,23 @@ const FlowChartTab = () => {
           data-rfinstance={JSON.stringify(nodes)}
         >
           <NodeEditMenu
-            selectedNode={selectedNode}
+            selectedNode={selectAllNodes ? null : selectedNode}
             unSelectedNodes={unSelectedNodes}
             manifestParams={manifestParams}
           />
 
-          <ReactFlow
```
JeffDotPng, line 237: (Related to my previous comment) You should put the ReactFlow instance back here and use the dummy component I mentioned, instead of creating a wrapper.

src/feature/flow_chart_panel/views/ControlBar.tsx
Jin1L: @JeffDotPng 
So instead of calling it like this:   
```
useKeyboardShortcut("ctrl", "p", () => onRun(nodes, edges));
  useKeyboardShortcut("ctrl", "c", cancelFC);
  useKeyboardShortcut("ctrl", "s", () => saveFile(nodes, edges));
```
You are suggesting that the callback function should be related to the respective buttons?

JeffDotPng: Just put the hook in the respective buttons, they already have to get `useFlowChartGraph` to get the necessary state, avoiding `ControlBar` constantly rerendering.

src/feature/flow_chart_panel/views/ControlBar.tsx
```diff
@@ -215,6 +217,7 @@ const LoadButton = () => {
...
```
JeffDotPng, line 220: This hook is still being called in `ControlBar`.. I would recommend extracting the cancel button into a component and calling it in there instead, along with putting the keyboard shortcut hook inside of there.

src/feature/flow_chart_panel/FlowChartTabView.tsx
```diff
@@ -229,11 +266,13 @@ const FlowChartTab = () => {
           data-rfinstance={JSON.stringify(nodes)}
         >
           <NodeEditMenu
-            selectedNode={selectedNode}
+            selectedNode={nodes.every((n) => n.selected) ? null : selectedNode}
```
JeffDotPng, line 269: Should probably be null if more than 1 node is selected instead of all of them

JeffDotPng (changes requested):

JeffDotPng (approved):
Looks good, lgtm

PR #501, opened 2023-05-31
fixed a small bug related to dragging

Orisu179:
Now when the user is able to scroll through the select section without dragging the entire menu


itsjoeoui (approved):
lgtm

PR #516, opened 2023-06-06
Add Node Cypress test case

Lans1975:


yassine-moumen:
## Use this URL to access Flojo-studio 

http://54.235.63.181:3000 
http://516.flojoy-test.com:3000 (comming soon) 

The instance will shutdown after one hour. 
If you want to start the instance after that use this API : 
https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-089354d30a25d73b6

check-dependencies
jackparmer, line 1: this should not be committed since it's a hot fix for your own machine

cypress/e2e/02_add_delete_node_page.cy.ts
```diff
@@ -0,0 +1,63 @@
...
+  beforeEach(() => {
+    // Open Eyes to start visual testing.
+    // Each test should open its own Eyes for its own snapshots.
+    cy.eyesOpen({
+      // The name of the application under test.
+      // All tests for the same app should share the same app name.
+      // Set this name wisely: Applitools features rely on a shared app name across tests.
+      appName: "studio",
+
+      // The name of the test case for the given application.
```
jackparmer, line 14: i don't think that we need to keep all of these comments that were copy/pasted from the Applitools cypress tutorial?

cypress/e2e/02_add_delete_node_page.cy.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    cy.get('[data-testid="add-node-button"]').click();
+
+    cy.eyesCheckWindow({
+      tag: "dark flow page with add node sidebar",
+      target: "window",
+      fully: true,
+    });
+    cy.xpath("//div[contains(text(), 'Loaders')]").click();
```
jackparmer, line 41: Would be good to have short 1-line comments about what each of these actions are doing

requirements.txt
```diff
@@ -5,7 +5,7 @@ django-cors-headers==3.13.0
 flojoy
 matplotlib==3.5.2
 networkx==2.8.4
-numpy==1.23.0
+numpy==1.24.3
```
jackparmer, line 8: should this be updated for everyone? if so, please make a note on why?

jackparmer (changes requested):

requirements.txt
Lans1975: Not sure

cypress/e2e/02_add_delete_node_page.cy.ts
Lans1975: Will add

cypress/e2e/02_add_delete_node_page.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+    cy.get('[data-testid="sidebar-close"]').click();
+    cy.eyesCheckWindow({
+      tag: "dark flow page with node loader",
+      target: "window",
+      fully: true,
+    });
+    // Click on added container LOADER
+    cy.xpath("//div[contains(text(), 'LOADER')]").click();
+    //Delete node LOADER
+    cy.get('.tabler-icon-x[width=\'24\']').click();
```
JeffDotPng, line 41: Can you add a check that the "LOADER" node is actually gone from the document?

package.json
```diff
@@ -145,6 +145,7 @@
     "@vitejs/plugin-react-swc": "^3.3.0",
     "concurrently": "^8.0.1",
     "cypress": "^12.11.0",
+    "cypress-xpath": "^2.0.1",
```
JeffDotPng, line 148: Just a warning that this package recently became deprecated, but it should be fine to keep using.

cypress/e2e/02_add_delete_node_page.cy.ts
Lans1975: Done

package.json
Lans1975: We can remove this package when we will have data-test for the component involved in Cypress automation.

jackparmer (approved):

PR #530, opened 2023-06-09
[DO-NOT-MERGE] Improve watch py wait loop

dev-nested:
Currently we can see a visible lag when a node finishes and the next node starts. This PR tries to improve the time it takes to switch execution between nodes.

Currently the job scheduler schedules jobs in batches. If it had scheduled 3 jobs in the previous batch, it'll first wait for all 3 jobs to finish. Then it'll find out which jobs have become newly ready and schedule them in a batch. 

This PR changes the scheduler to not wait for all the jobs in a batch to start a new batch of job. Instead, now it runs in a fixed periodical loop. On each iteration it checks if there are new jobs to be scheduled, and then it checks if any of the running jobs has finished.

Tested with the default example app.

### Additional notes
- Has performance improved?
  - Yes. possibly but not visibly. Certainly not slower. Theoritically, if we use a small period for the wait loop, it will work as fast as it gets with each node being run in separate rq-worker call.
- What do we do with this PR?
  - I recommend to keep it open and unmerged for now, and implement another approach we have planned. With that new approach, all regular python functions will be executed in the scheduler itself (or one dedicated rq worker). This will achieve minimum lag between nodes because it'll then essentially be same as calling two functions in same program (thread/memory space). There will be an option to mark a node as needing independent rq worker which might be required for long running nodes.
  - See this related issue https://github.com/flojoy-io/studio/pull/489 which implements and compares event driven approach.

### DIFF
--- a/PYTHON/WATCH/watch.py
+++ b/PYTHON/WATCH/watch.py
@@ -47,10 +47,8 @@ class FlowScheduler:
         self.topology.print_graph()
         self.topology.collect_ready_jobs()
 
-        num_times_waited_for_new_jobs = 0
-        wait_time_for_new_jobs = 0.1
-        wait_time_multiplier = 2
-        max_wait_time = 10
+        wait_time = 0.1
+        running_jobs = []
 
         while not self.topology.finished():
             print("\nnext wave")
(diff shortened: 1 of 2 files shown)


yassine-moumen:

http://3.84.28.113:3000 
http://530.flojoy-test.com:3000 

https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-0236491031451412b

jackparmer:
@yassine-moumen 

> http://3.84.28.113:3000/
> http://530.flojoy-test.com:3000/

which of these 2 links is which? i assume that the 1st one is windows and the 2nd one is 'nix? can you please label these links with which operating system they are?

jackparmer:
Also @mdasifhasan reports that the default app runs locally for him, but it does not run for either of the preview builds. This might be a good test case to explore

smahmed776:
We got rid of Django and Watch.py files so this PR seems redundant now. closing...

PR #541, opened 2023-06-10
Debugging python nodes in vscode

dev-nested:
Debugging nodes is hard right now. Currently we rely on console printing the values which is very tedious. Also with all the commands now run in a single tab, we don't see printouts from node functions reliably. This PR adds debugging support for IDE. It is tested with vscode for now. It might work with other IDEs that supports debugging through [debugpy].(https://github.com/microsoft/debugpy)

Used the  [debugpy](https://github.com/microsoft/debugpy) library to support debugging in vscode. This PR adds a separate rq worker implementation which starts the debugger. The debugger can be configured through environment variables. It currently supports three variables.
1. `DEBUG`: This turns debugging mode on for node python functions. Acceptable values are `True`, `False`. By default debugging is disabled.
1. `DEBUG_PORT`: The port where the debugger should listen to. Accepts a valid port number. Default is `5678`.
1. `DEBUG_WAIT`: Whether the debugger should wait for a client (vscode) to connect. Acceptable values are `True`, `False`. Default is false.

To debug, following steps are required:
1. Run flojoy with the `DEBUG` env variable set, for example:
```
DEBUG=True bash flojoy
```
2. In vscode, open the debug tab, run `Start Debugging`

Screenshot 2023-06-10 at 1 19 54 PM

3. Set breakpoints where you want to. You can set breakpoint inside pip modules as well (e.g. `@flojoy` decorator)

Screenshot 2023-06-10 at 1 21 14 PM

That's it. Now as you run the flow chart in frontend, VS code will steal focus to let you debug.

Tested with the default example app in local machine following the above steps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ worker-manager/node_modules
 
 /**/.idea/*
 .vscode/*
+!.vscode/launch.json
 
 src/data/*.json
 src/feature/flow_chart_panel/manifest/pythonFunctions.json
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Python: Attach",
+      "type": "python",
+      "request": "attach",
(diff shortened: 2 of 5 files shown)


yassine-moumen:

http://54.205.241.173:3000 
http://541.flojoy-test.com:3000 

https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-07fd55bea8cfa579a

PR #545, opened 2023-06-11
Sci num py scrapper

trbritt:
This contains the added design / UI for the autogenerated nodes for the `scipy.stats`, `scipy.signal`, and `numpy.linalg` (more are to follow. Files are autogenerated using new functionality in `flojoy/python`, see https://github.com/flojoy-io/python/pull/38.


yassine-moumen:

http://52.70.218.153:3000 
http://545.flojoy-test.com:3000 

https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-0640c5aaa99ca088e

trbritt:
This PR makes the following additions: added new keys for autogenerated nodes, adds custom styling for Numpy and Scipy nodes that includes their SVG logos to give accredidation, allows nodes to have `null` default values for parameters. Many numpy and scipy functions require more information to operate than a single input array, but information that seems irrelevant enough to keep them as parameters and not entire nodes to pass to the scipy/numpy node.

Issue #529, opened 2023-06-09
Improve how multiple inputs are passed to nodes


dev-nested:
There is a major issue with nodes with multiple inputs, it seems that the order of the inputs is not guaranteed, which causes problems when the order matters.

The node function signature should be like this

```
def SINE(inputs:dict[str, DataContainer], params:dict)-> Union(DataContainer, dict[str, DataContainer]):
```

Please also update node_styleguide.md for contributors (in the nodes repo)

Support for multiple outputs is covered in this issue https://github.com/flojoy-io/studio/issues/524

dev-nested:
Had a discussion with @JeffDotPng and @itsjoeoui for a better design on how multiple inputs are supported. Here's the new approach we decided to implement.

- The node functions will define all its inputs and params as regular python function parameters with proper type hint.
- All the function parameters that are of DataContainer type are inputs
- All the function parameters that are not DataContainer type are params
- For example, a node with two inputs a and b, and two parameters p and q, will be written as

```
def ADD(a: DataContainer, b: DataContainer, p: int, q: str):
  ...
  ...function body
  ...
```

The `@flojoy` decorator has to properly propagate the arguments to all the function parameters by reading the signature of the target function. It has to utilize the type hints to distinguish inputs from node params.

itsjoeoui:
Thanks Asif for documenting all these, can't wait for us to finish this one. 🚀

PR #556, opened 2023-06-13
#529 pass to an enqueued job the input names its previous jobs connects to

dev-nested:
### Why is this change needed?
Currently a node receives its inputs as an array. It's been reported that the ordering of inputs do not stay fixed. We need a way for nodes to identify inputs by their name. 

This change sends to the `@flojoy` decorator the input names that a node's previous jobs connect to. 

See related PR https://github.com/flojoy-io/python/pull/40

### What is the solution implemented?
When enqueuing an rq job, the scheduler (`Watch.py`) now sends a dictionary of previous jobs where each job item contains the previous job id and the input name it connects to. Before this change the scheduler used to send only the previous job ids as a list.

It is backward compatible and can be merged independently. It still sends the previous job ids, in addition now sends a dictionary with previous jobs information.

I ran the default app and checked that the `@flojoy` decorator now receives the input names. Here's a screenshot of a debugging session for the ADD node inspecting the `**kwargs` value of the `@flojoy` decorator. It shows that previous_jobs received the expected dictionary.

Here's a screenshot:
Screenshot 2023-06-15 at 11 17 00 PM

**Issue:** https://github.com/flojoy-io/studio/issues/529

### DIFF
--- a/PYTHON/WATCH/watch.py
+++ b/PYTHON/WATCH/watch.py
@@ -139,13 +139,19 @@ class FlowScheduler:
             except Exception:
                 pass
 
-        dependencies = self.topology.get_job_dependencies(job_id, original=True)
+        previous_job_ids = self.topology.get_job_dependencies(job_id, original=True)
+        previous_jobs = self.topology.get_job_dependencies_with_label(
+            job_id, original=True
+        )
 
         print(
             " enqueue job:",
             self.topology.get_label(job_id),
             "dependencies:",
-            [self.topology.get_label(dep_id, original=True) for dep_id in dependencies],
+            [
+                self.topology.get_label(dep_id, original=True)
+                for dep_id in previous_job_ids
+            ],
         )
 
         self.job_service.enqueue_job(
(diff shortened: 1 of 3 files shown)


yassine-moumen:

Windows installation: http://556_windows.flojoy-test.com:3000 
Linux installation: http://556_linux.flojoy-test.com:3000 

It is important to know that the instance will automatically shut down after 4 hours to conserve resources. If you wish to start the instance again after it shuts down (or restart it to pickup new changes), you can use this API : 

Windows: https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-0f33f4478c287f582&domain=556_windows 

Linux: https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-04b610f9b96508791&domain=556_linux

dev-nested:
@yassineMoumen can we have it tested in the live test environment?

itsjoeoui (approved):

PR #618, opened 2023-06-22
Fixed handlecomponent to render input and outputs correctly

smahmed776:
### Changes:
- Refactored HandleComponent to render all inputs defined in the node manifest file if there are no inputs defined in the manifest file then only render default one source handle and the same for the target handle. 

Previously, the HandleComponent was rendering a default input for all of the nodes regardless of inputs defined in the node manifest file. For example, if there is two input defined in the node manifest file it'd render three source handle and the same for the target handle.
### Before
Screenshot (79)

### After
Screenshot (78)


dstrande (changes requested):
When trying to connect to the bottom connection, it automatically jumps to the top connection.

image

PR #643, opened 2023-07-01
Loop and Conditional working

smahmed776:


yassine-moumen:

Windows installation: http://643_windows.flojoy-test.com:3000 
Linux installation: http://643_linux.flojoy-test.com:3000 

It is important to know that the instance will automatically shut down after 4 hours to conserve resources. If you wish to start the instance again after it shuts down (or restart it to pickup new changes), you can use this API : 

Windows: https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-0736ca7d1920f74b1&domain=643_windows 

Linux: https://kgpsoscpc4q52d4mtwuu7jhvie0qdiso.lambda-url.us-east-1.on.aws/start_instance?instance_id=i-0590d0e2e5bee39f7&domain=643_linux

PYTHON/manifest/manifest.py
```diff
@@ -79,7 +95,9 @@ def func(name: str, input_type: str):
                 )
 
             params[name] = {
-                "type": get_full_type_name(param_type),
+                "type": "node_reference"
```
JeffDotPng, line 98: We should find a better way to encode the fact that a parameter is a node reference instead of hardcoding nodes that it should work for.

PYTHON/manifest/manifest.py
```diff
@@ -91,6 +109,8 @@ def func(name: str, input_type: str):
     outputs = []
 
     create_output = create_io(outputs)
+    if not sig.return_annotation:
```
JeffDotPng, line 112: Shouldn't all nodes have a return annotation?

PYTHON/manifest/manifest.py
```diff
@@ -113,28 +133,63 @@ def func(name: str, input_type: str):
...
+                    f"consisting of only DataContainers as fields, got {return_type} for {node_name}"
+                )
+            try:
+                output_type = (
+                    union_type_str(annotation)
+                    if is_union(annotation)
+                    else annotation.__name__
+                )
+            except Exception:
+                if node_name in SPECIAL_NODES:
```
JeffDotPng, line 156: Should do this check first and return early instead of using exception base control flow

PYTHON/manifest/manifest.py
```diff
@@ -113,28 +133,63 @@ def func(name: str, input_type: str):
...
+                    output_type = "Any"
```
JeffDotPng, line 157: I should probably change 'any' to be uppercase like this just to be consistent

PYTHON/manifest/manifest.py
```diff
@@ -113,28 +133,63 @@ def func(name: str, input_type: str):
...
+                else:
+                    raise
+
+            create_output(attribute, output_type)
+
     elif is_dataclass(return_type):
         for field in fields(return_type):
-            if not issubclass(field.type, DataContainer):
+            if (not inspect.isclass(field.type) and node_name not in SPECIAL_NODES) or (
```
JeffDotPng, line 165: You should probably refactor this complicated condition into a function since it's used twice

PYTHON/manifest/manifest.py
```diff
@@ -113,28 +133,63 @@ def func(name: str, input_type: str):
...
-                else field.type.__name__
-            )
+            try:
+                output_type = (
+                    union_type_str(field.type)
+                    if is_union(field.type)
+                    else field.type.__name__
+                )
+            except Exception:
+                if node_name in SPECIAL_NODES:
```
JeffDotPng, line 180: Same thing as earlier

PYTHON/dao/redis_dao.py
```diff
@@ -23,65 +24,64 @@ def get_instance():
...
+    def get_set_list(self, key: str):
         return self.r.smembers(key)
 
     def serialize_np(self, np_array: np.ndarray):
         return np_array.ravel().tostring()
 
-    def desirialize_np(self, encoded: str, np_meta_data: dict):
-        if encoded is None:
-            return []
+    def desirialize_np(self, encoded: bytes, np_meta_data: dict[str, str]):
```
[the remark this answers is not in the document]
smahmed776: Fixed

PYTHON/manifest/manifest.py
smahmed776: Introduced ParameterTypes class to handle this.

PYTHON/manifest/manifest.py
smahmed776: A node that doesn't have a source handle should not contain return annotation and a node that doesn't have a target handle should not contain any parameter with DC type annotation, this is my point of view. Following this role we don't have to exclude any nodes when generating output in the manifest.

captain/models/topology.py
```diff
@@ -78,12 +80,10 @@ async def run_job(self, job_id):
...
             func = getattr(module, cmd)
 
-        dependencies = self.get_job_dependencies(job_id)
-        previous_jobs_with_label = self.get_job_dependencies_with_label(job_id)
-        logger.debug(f"labelled previous jobs: {previous_jobs_with_label}")
+        previous_jobs = self.get_job_dependencies_with_label(job_id, original=True)
 
         logger.debug(
-            f" enqueue job: {self.get_label(job_id)}, dependencies: {[self.get_label(dep_id, original=True) for dep_id in dependencies]}"
+            f" enqueue job: {self.get_label(job_id)}, dependencies: {[self.get_label(dep_id.get('job_id'), original=True) for dep_id in previous_jobs]}"
```
JeffDotPng, line 86: Should probably rename `dep_id` to `dep`

JeffDotPng (approved):

PR #685, opened 2023-07-14
Boolean switch

Orisu179:
Updated the boolean field from checkbox:
image
Into a switch: 
image
Light mode: 
image


src/assets/PlayBtnIconSVG.tsx
Orisu179, line 1: Unused line so I figured I can get rid of it

PR #711, opened 2023-07-19
Get rid of GOTO and END nodes

smahmed776:


captain/models/topology.py
```diff
@@ -57,14 +61,15 @@ async def run(self):
...
             time.sleep(self.node_delay)
 
     def collect_ready_jobs(self):
         next_jobs: list[str] = []
-        for job_id in cast(list[str], self.original_graph.nodes):
+        for job_id in cast(list[str], self.working_graph.nodes):
             if (
                 job_id not in self.finished_jobs
-                and self.original_graph.in_degree(job_id) == 0
+                and self.working_graph.in_degree(job_id) == 0
```
izi-on, line 72: I set this to original_graph to prevent future devs from using this on the working_graph to collect nodes with in-degree 0 during flowchart run.

captain/utils/flowchart_utils.py
```diff
@@ -141,9 +141,17 @@ async def prepare_jobs_and_run_fc(request: PostWFC, manager: Manager):
...
         manager.end_worker_threads()
         clear_memory()
+        if is_finished:
+            socket_msg = WorkerJobResponse(
+                jobset_id=request.jobsetId,
+                sys_status=STATUS_CODES["STANDBY"],
+                failed_nodes=[],
+                running_node="",
+            )
+            asyncio.create_task(manager.ws.broadcast(socket_msg))
```
izi-on, line 154: would rather have anything related to message broadcasting done in the broadcast utils if possible :)

captain/types/worker.py
```diff
@@ -40,16 +40,16 @@ def __init__(
...
         dict_item: dict[str, Any] = {},
     ):
         self["jobsetId"] = jobset_id
         if sys_status:
             self["SYSTEM_STATUS"] = sys_status
         self["type"] = "worker_response"
-        self["FAILED_NODES"] = failed_nodes
-        self["RUNNING_NODES"] = running_nodes
+        self["FAILED_NODES"] = failed_nodes or []
+        self["RUNNING_NODE"] = running_node
```
izi-on, line 52: fyi planning on supporting multiple node highlighting in the future, so it should be kept to running_nodes

captain/utils/flowchart_utils.py
izi-on, line 154: this might seem annoying but anything related to broadcasting should be in the broadcast utils, slowly working on putting everything there :)

captain/routes/flowchart.py
```diff
@@ -65,7 +65,7 @@ async def write_and_run_flowchart(request: PostWFC):
 async def worker_response(
     request: Request,
 ):  # TODO figure out way to use Pydantic model, for now use type Request otherwise does not work????
-    if manager.running_topology is None or manager.running_topology.is_cancelled():
+    if manager.running_topology is None or manager.running_topology.cancelled:
```
izi-on, line 68: Seems annoying, but better to have a method for this, we want to follow the paradigm where we treat topology like a separate object/model and not something we can access its attributes easily.

captain/models/topology.py
```diff
@@ -39,11 +38,16 @@ def __init__(
         self.node_delay = node_delay
         self.max_runtime = max_runtime
         self.finished_jobs: set[str] = set()
+        self.queued_jobs: set[str] = set()
```
izi-on, line 41: queued_jobs is useless I believe

captain/models/topology.py
```diff
@@ -117,11 +122,14 @@ async def run_job(self, job_id: str):
                 previous_jobs=previous_jobs,
             )
         )
+        if node["cmd"] == "LOOP":
+            self.loop_nodes.append(job_id)
```
izi-on, line 126: There is an opportunity here to make the code more readable, maybe have a function is_loop_node(node_id) then put the node = cast(... ad the if condition there, makes for better code :) your call

captain/models/topology.py
```diff
@@ -193,24 +191,34 @@ def process_job_result(
         if not next_directions:
             next_directions = self.get_outputs(job_id)
 
-        logger.debug(f"out_edges: {self.get_outputs(job_id)}")
+        logger.debug(f"out_edges to follow: {next_directions}")
 
         for direction_ in next_directions:
             direction = direction_.lower()
+            if direction == "end" and self.loop_nodes:
```
izi-on, line 198: a little confused by this here, weren't we getting rid of end?

captain/models/topology.py
```diff
@@ -193,24 +191,34 @@ def process_job_result(
...
+        ):
+            if not self.loop_nodes:
+                self.is_finished = True
+                logger.debug(
+                    f"FLOWCHART TOOK {time.time() - self.time_start} SECONDS TO COMPLETE"
+                )
+                self.cancel()
+                return
+            else:
+                nodes_to_add.append(self.loop_nodes[-1])
```
izi-on, line 216: clever!

captain/utils/flowchart_utils.py
```diff
@@ -141,9 +141,17 @@ async def prepare_jobs_and_run_fc(request: PostWFC, manager: Manager):
...
 
     # clean up before next run
     clean_up_function()
```
izi-on, line 157: overall very good ! I am impressed you managed to remove GOTO in such a short amount of code :0

izi-on (changes requested):
just reply to my comments and if you want make the changes, just some recommendations that's it :) very good stuff

captain/models/topology.py
smahmed776: ok, it works as this only executes once. keeping the old change.

captain/types/worker.py
smahmed776: Keeping this for now as we have to adjust the front end as well to support multiple nodes running at a time. we can implement this in the near future when needed as a separate task.

captain/routes/flowchart.py
smahmed776: Yes, makes sense. bringing that method back.

captain/models/topology.py
smahmed776: Adjusted as per your recommendation, thanks

izi-on (approved):
lgtm!

PR #713, opened 2023-07-19
change api key feature to support list of keys

youngsun4786:
The goal is to generalize the API key feature so that users, developers do not have to face the discomfort of making PRs for including additional 3rd party services.


src/hooks/useFlowChartState.ts
```diff
@@ -49,10 +54,8 @@ const rfInstanceAtom = atomWithImmer<
 const editModeAtom = atomWithImmer<boolean>(false);
 const expandModeAtom = atomWithImmer<boolean>(false);
 const apiKeyAtom = atomWithImmer<string>("");
-const s3NameAtom = atomWithImmer<string>("");
-const s3AccessKeyAtom = atomWithImmer<string>("");
-const s3SecretKeyAtom = atomWithImmer<string>("");
-const s3ContainerAtom = atomWithImmer<boolean>(false);
+const apiValueAtom = atomWithImmer<string>("");
+const credentialsAtom = atomWithImmer<any>({});
```
JeffDotPng, line 58: type should probably be `{ [key: string]: string }`

src/hooks/useFlowChartState.ts
```diff
@@ -35,6 +35,11 @@ export interface CtlManifestType {
   layout: ReactGridLayout.Layout;
 }
 
+export type API_TYPE = {
```
JeffDotPng, line 38: Why is this in here? it's only used in one place, `FlowChartServices.ts`.
Also, type names should be `PascalCase` not `ALL_CAPS`

src/services/FlowChartServices.ts
```diff
@@ -14,28 +14,16 @@ const flowKey = "flow-joy";
...
-
 export function saveFlowChartToLocalStorage(rfInstance?: ReactFlowJsonObject) {
   if (rfInstance) {
     const flowObj = rfInstance;
     localforage.setItem(flowKey, flowObj);
   }
 }
 
-export const sendApiKeyToDjango = async (body: API_TYPE, endpoint: string) => {
+export const sendApiKeyToFastAPI = async (body: API_TYPE) => {
```
JeffDotPng, line 24: Can you just call this function `postEnvironmentVariable`

src/themeContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const toggleTheme = useCallback(() => {
+    const toggleTheme = theme === "light" ? "dark" : "light";
+    setTheme(toggleTheme);
+    window.localStorage.setItem("theme", toggleTheme);
+  }, [theme]);
+
+  useEffect(() => {
+    const localTheme = window.localStorage.getItem("theme");
+    localTheme && setTheme(localTheme);
```
JeffDotPng, line 35: Can you just use an if statement for this line, not a big fan of this

src/feature/flow_chart_panel/views/APIKeyModal.tsx
```diff
@@ -1,337 +1,155 @@
 import FamilyHistoryIconSvg from "@src/assets/FamilyHistoryIconSVG";
```
JeffDotPng, line 1: Not related just to this file but we should probably rename all of these "APIKey(\*)" related things to "EnvVar(\*)", it's more descriptive since it's just any environment variable and not just API keys

src/feature/flow_chart_panel/views/APIKeyModal.tsx
```diff
@@ -1,337 +1,155 @@
...
+              type="password"
+              value={apiValue || ""}
+              className="mt-1 w-72 text-black shadow-sm dark:bg-neutral-800 dark:text-white sm:text-sm "
+              onPaste={splitOnCopy}
+              onChange={handleApiValueChange}
+            />
+          </div>
+        </div>
+        <DialogFooter className="px-3">
+          <Button onClick={handleSendAPI}>Submit</Button>
```
JeffDotPng, line 129: Probably better if the button text is 'Add'

PR #724, opened 2023-07-24
Update styles

JeffDotPng:
- Restyled sidebar a bit
- Made some nodes more responsive with respect to size
- Nodes now inherit their style from their parent if they do not have a valid type
- Refactored handle component logic for logic nodes
- Added context menu for inspect element in electron
image

Corresponding PR to fix some node types: https://github.com/flojoy-io/nodes/pull/192


electron/main/index.ts
```diff
@@ -35,13 +39,21 @@ if (!app.requestSingleInstanceLock()) {
...
+});
+
 let win: BrowserWindow | null = null;
 // Here, you can also use other preload
 const preload = join(__dirname, "../preload/index.js");
 const url = process.env.VITE_DEV_SERVER_URL;
 const indexHtml = join(process.env.DIST, "index.html");
 
 async function createWindow() {
+  installExtension(REACT_DEVELOPER_TOOLS)
```
itsjoeoui, line 53: Can u make it install only for dev but not prod?

tailwind.config.js
```diff
@@ -15,6 +15,16 @@ module.exports = {
     "bg-accent2",
     "bg-accent3",
     "bg-accent4",
+    "bg-accent1/5",
```
itsjoeoui, line 18: what is this for again?

itsjoeoui (changes requested):

tailwind.config.js
JeffDotPng: you can't use a tailwind class dynamically (passing as a prop or whatever) unless you use it somewhere else statically, because it won't be included in the css bundle otherwise
this ensures that the class is always included in the css bundle (it's kind of annoying but it is what it is)

electron/main/index.ts
JeffDotPng: turns out installing react dev tools in electron is just broken as a whole right now, i got rid of the package

itsjoeoui: lol classic react ecosystems

itsjoeoui (approved):

PR #725, opened 2023-07-24
[nodes] Fix node types (#192)

smahmed776:
- Moved all custom node components to [flojoy package](https://github.com/flojoy-io/types)


JeffDotPng (approved):
lgtm

Issue #194, opened 2023-01-31
UI - connect nodes via not shortest line distance


trbritt:
In order to design an active feedback loop, very often nodes at the end of the chain will need to be connected to a comparator at the beginning. As of right now, the nodes are connected by the shortest curved path possible, but for these applications this creates a connection that clutters the rest of the nodes (see below).

Is it possible to intentionally increase the length of the curve connecting the nodes to move out of the way, maybe drop completely below the nodes like a 'U' shape?

![image](https://user-images.githubusercontent.com/53545754/215900187-3ecbb184-a8e6-4c01-9920-a70464a94a61.png)

PR #729, opened 2023-07-25
[nodes] Merge pull request #194 from flojoy-io/reda-fix-tmin-test

smahmed776:


captain/models/topology.py
```diff
@@ -196,11 +196,13 @@ def process_job_result(
...
         logger.debug(f"out_edges to follow: {next_directions}")
 
-        for direction_ in next_directions:
+        for i, direction_ in enumerate(next_directions):
             direction = direction_.lower()
             if direction == "end" and self.loop_nodes:
                 self.loop_nodes.pop()
             self.mark_job_success(job_id, next_nodes_from_dependencies, direction)
+            if i == len(next_directions) - 1:
+                self.working_graph.remove_node(job_id)
```
izi-on, line 205: Maybe I am wrong, but wouldn't it be better to remove the edges out of which it flows instead of the entire node?

izi-on (changes requested):

captain/models/topology.py
smahmed776: This would create a conflict as there will be several nodes with out_degree == 0. So to be safe I considered removing node completely.

izi-on (approved):
lgtm!

PR #747, opened 2023-08-03
App Gallery Search bar

Orisu179:


src/feature/flow_chart_panel/views/AppGalleryModal.tsx
```diff
@@ -43,7 +58,7 @@ export const AppGalleryModal = ({
       const filtered = raw.filter(
         (obj) => obj["type"] === "dir" && !ignoreDir.includes(obj["name"])
       );
-      setSelect(
+      setSelectFields(
         filtered.map((obj) => (
           <SelectItem key={obj.sha} value={obj.url}>
             {obj.name}
```
JeffDotPng, line 64: Instead of storing an array of components as state, I think it would be better to just store the data then map it to the `SelectItems` in the `SelectContent`

JeffDotPng (changes requested):
Pretty cool feature! Needs a few UX changes though (see the comment related to search data)
Also maybe there could be some text indicating what this is for VS what the app gallery is for (may need some layout changes, if you can't figure something out I can do it at some point), something like "Search for node-specific example apps..." and then "or browse our gallery of apps for specific tasks..."

src/feature/flow_chart_panel/views/AppGalleryModal.tsx
```diff
@@ -22,18 +22,33 @@ import {
...
+}
 
 export const AppGalleryModal = ({
   isGalleryOpen,
   setIsGalleryOpen,
 }: AppGalleryModalProps) => {
-  const [selectFields, setSelect] = useState([]);
-  const [data, setData] = useState<object[]>([]);
+  const [selectFields, setSelectFields] = useState([]);
+  const [searchData, setSearchData] = useState<GithubJSON[]>([]);
```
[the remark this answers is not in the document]
Orisu179: I was able to do this, but it seems like the search component I was using would freeze when the file headings are too much (the limit is around 15 ish?). Is it ok if I disable the search bar at first and contain a message like "please select a category" instead? Otherwise, I would need to do some big refactors to make it work...

src/feature/flow_chart_panel/views/AppGalleryModal.tsx
```diff
@@ -18,62 +18,65 @@ import {
...
-            {obj.name}
-          </SelectItem>
-        ))
-      );
-    };
-    fetchData().catch(console.error);
-  }, []);
+  const [selectFields, setSelectFields] = useState<GithubJSON[]>([]);
+  const [searchData, setSearchData] = useState<GithubJSON[]>([]);
+  const [searchDisabled, setSearchDisabled] = useState<boolean>(true);
```
JeffDotPng, line 51: This should be set to true again when you reopen the modal (just use a useEffect)

src/feature/flow_chart_panel/views/AppGalleryModal.tsx
```diff
@@ -85,19 +88,33 @@ export const AppGalleryModal = ({
...
             <div className="ml-6 text-3xl">App Gallery</div>
-            <div className="ml-72 flex gap-5">
-              <AppGallerySearch items={data} />
-              <div className="mt-0.5 pt-1">
-                <Select>
-                  <SelectTrigger>Node Category</SelectTrigger>
+            <div className="flex basis-2/5 gap-3">
+              <div className="w-1/3 pt-1">
+                <Select onValueChange={onValueChange}>
+                  <SelectTrigger className="text-sm">Category</SelectTrigger>
```
JeffDotPng, line 97: Probably good to make the text inside change to the currently selected category

JeffDotPng (changes requested):

JeffDotPng:
Closing in favor of #769

PR #751, opened 2023-08-03
make the app gallery modal close when we select an example

Jin1L:


src/feature/flow_chart_panel/views/AppGalleryModal.tsx
```diff
@@ -18,11 +18,12 @@ import {
   SelectItem,
   SelectTrigger,
 } from "@/components/ui/select";
-import { DialogClose } from "@radix-ui/react-dialog";
+import { useFlowChartState } from "@src/hooks/useFlowChartState";
 
 const subjectKeyList = ["fundamentals", "AI", "IO", "DSP"];
 const ignoreDir = [".github", "MANIFEST"];
 export const AppGalleryModal = () => {
+  const { isGalleryOpen, setIsGalleryOpen } = useFlowChartState();
```
JeffDotPng, line 26: There is no reason for this to be global state, you can easily make this a `useState` and pass it down as a prop from `FlowChartTabView` instead.

JeffDotPng (changes requested):

src/feature/flow_chart_panel/views/AppGalleryModal.tsx
Jin1L: But I also use that inside the ```AppGalleryElement.tsx``` file to close down the model when we click an example

JeffDotPng: You can pass it down through `AppGalleryLayout`'s props and then pass it down again to `AppGalleryElement`.

JeffDotPng (approved):

PR #758, opened 2023-08-04
Play button disabled when no nodes are on canvas

Jin1L:


src/feature/flow_chart_panel/components/PlayBtn.tsx
```diff
@@ -23,7 +23,12 @@ const PlayBtn = ({ onPlay }: PlayBtnProps) => {
   useKeyboardShortcut("meta", "p", () => onPlay(nodes, edges));
 
   return (
-    <Button size="sm" variant="outline" onClick={handleClick}>
+    <Button
+      size="sm"
+      variant="outline"
+      onClick={handleClick}
+      disabled={!nodes.length}
```
JeffDotPng, line 30: I think `nodes.length === 0` is more clear

youngsun4786: sounds good will change it back

JeffDotPng (approved):

PR #761, opened 2023-08-04
ensure editing keys get updated right away for env vars

youngsun4786:


src/feature/common/DarkModeToggle.tsx
```diff
@@ -19,10 +20,7 @@ export const DarkModeToggle = () => {
     <Button
       size="icon"
       variant="ghost"
-      onClick={() => {
```
JeffDotPng, line 22: don't we still need this to not break all of the mantine stuff?

Jin1L: We put the maintine stuff inside toggleTheme function to make it not break

JeffDotPng: Oh oops i didn't see that

youngsun4786: yeah just so that onClick is only taking one single prop. If this is bad practice or might cause issues later, I'll revert back the changes.

JeffDotPng (approved):

PR #783, opened 2023-08-13
Check for update

itsjoeoui:
this is a bit hacky, but may be the best solution without a proper electron updater

The check for update will first check if there is a new commit on the current branch, if there is, a toast will be shown with an update button. Pressing the update button will simply do a git pull

### DIFF
--- a/src/feature/flow_chart_panel/views/ControlBar.tsx
+++ b/src/feature/flow_chart_panel/views/ControlBar.tsx
@@ -34,8 +34,8 @@ import {
   MenubarMenu,
   MenubarTrigger,
 } from "@src/components/ui/menubar";
-import { Button } from "@src/components/ui/button";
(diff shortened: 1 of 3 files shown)


src/feature/flow_chart_panel/views/ControlBar.tsx
```diff
@@ -273,6 +275,29 @@ const ControlBar = () => {
...
+          label: "Update",
+          onClick: async () => {
+            await fetch(`${API_URI}/update/`, {
+              method: "POST",
+            });
+          },
+        },
+      });
+    } else {
+      toast("Your Flojoy Studio is update-to-date");
```
JeffDotPng, line 297: 'up to date' instead of 'update-to-date'

JeffDotPng (approved):

JeffDotPng (approved):

PR #828, opened 2023-08-24
Enable/disable electron window from startup script

smahmed776:


luiztauffer (approved):

luiztauffer:
@smahmed776 thanks a bunch! this looks good to me

PR #830, opened 2023-08-25
Properly package the app installer

smahmed776:


PYTHON/generate_manifest.py
```diff
@@ -102,7 +102,9 @@ def browse_directories(dir_path: str, cur_type: Optional[str] = None):
             __generated_nodes.append(n_file_name)
         except Exception as e:
             print(
-                "❌ Failed to generate manifest from ",
+                "❌ Failed to generate manifest from ".encode("ascii", "replace").decode(
```
JeffDotPng, line 105: Why does this need ascii encoding? Does UTF-8 break things? If it's the emojis breaking we could probably just remove them

electron/api/index.ts
```diff
@@ -5,4 +5,18 @@ export default {
...
+        ipcRenderer.send(channel, data);
+      }
+    },
+    receive: (channel:string, func:(...arg:string[]) => void) => {
+      const validChannels = ['backend'];
+      if (validChannels.includes(channel)) {
+        // Deliberately strip event as it includes `sender`
+        ipcRenderer.on(channel, (event, ...args:string[]) => func(...args));
+      }
+    }
```
JeffDotPng, line 21: Why do this + defining `validChannels` when you could just wrap the channels like with `setUnsavedChanges`?

electron/main/backend.ts
```diff
@@ -0,0 +1,85 @@
+import * as childProcess from "child_process";
+import { join, resolve } from "path";
+import treeKill from "tree-kill";
+
+const sendBackendLogToStudio = (win: Electron.BrowserWindow, data: string) => {
```
JeffDotPng, line 5: `Electron.BrowserWindow` gives an error, need to `import { BrowserWindow } from 'electron'`

electron/main/backend.ts
```diff
@@ -0,0 +1,85 @@
...
+      sendBackendLogToStudio(win, data);
+      if (data.toString().includes(matchText)) {
+        sendBackendLogToStudio(win, "backend initialized successfully!");
+        resolve({ script });
+      }
+    });
+    script.addListener("exit", (code) => {
+      sendBackendLogToStudio(
+        win,
+        "Error: Failed to initialize backend try re lunching app!",
```
JeffDotPng, line 37:
```suggestion
        "Error: Failed to initialize backend, please try relaunching the app!",
```

electron/main/backend.ts
```diff
@@ -0,0 +1,85 @@
...
+        resolve({ script });
+      }
+    });
+    script.stderr?.on("data", function (data) {
+      sendBackendLogToStudio(win, data);
+      if (data.toString().includes(matchText)) {
+        sendBackendLogToStudio(win, "backend initialized successfully!");
+        resolve({ script });
+      }
+    });
```
JeffDotPng, line 33: The function name makes it seem like it's supposed to be a general function for running any command, but the logging from `win.webContents.send` show that it's only supposed to be used for backend initialization?

electron/main/backend.ts
```diff
@@ -0,0 +1,85 @@
...
+import { join, resolve } from "path";
+import treeKill from "tree-kill";
+
+const sendBackendLogToStudio = (win: Electron.BrowserWindow, data: string) => {
+  if (global.initializingBackend) {
+    win.webContents.send("backend", data);
+  }
+};
+
+export const runCommand = (
```
JeffDotPng, line 11: Can you use async/await instead of using `new Promise`?

smahmed776:
Changes moved to #854

PR #855, opened 2023-08-31
Renamed types in ManifestLoader.ts and used type guards in SideBarNode

marvinirwin:
I also added added `.nyc` folder to gitignore, because I guess i ran a command which filled that folder.

Feel free to absolutely assault me in code review.  Like no holds barred.  

The types were initially written with meaningless names, so I used regular tree words to name them instead, and also used type guards to remove the assertion in the SideBarNode component.


itsjoeoui (approved):
lgtm

src/feature/common/Sidebar/SidebarNode.tsx
```diff
@@ -1,7 +1,14 @@
+/* eslint-disable */
```
JeffDotPng, line 1: Can get rid of this along with unused imports

src/feature/common/Sidebar/SidebarNode.tsx
```diff
@@ -86,12 +89,19 @@ const SidebarNode = ({
...
-  if (node.name === "ROOT") {
+  debugger;
+  if (isRoot(node)) {
     return (
       <div>
         {node.children.map((c) => {
           // Actually needs to be called as a function to achieve depth-first traversal,
           // otherwise React lazily evaluates it and doesn't recurse immediately, resulting in breadth-first traversal.
+          let category, autogeneratedCategory, icon;
+          if (typeof c.key === "string") {
```
JeffDotPng, line 100: I think `c.key !== undefined` is clearer

src/feature/common/Sidebar/SidebarNode.tsx
```diff
@@ -86,12 +89,19 @@ const SidebarNode = ({
...
 }: SidebarNodeProps) => {
-  if (node.name === "ROOT") {
+  debugger;
+  if (isRoot(node)) {
     return (
       <div>
         {node.children.map((c) => {
           // Actually needs to be called as a function to achieve depth-first traversal,
           // otherwise React lazily evaluates it and doesn't recurse immediately, resulting in breadth-first traversal.
+          let category, autogeneratedCategory, icon;
```
JeffDotPng, line 99: Maybe add some explicit types for these? (`string | undefined` etc.) Editor gives warnings about implicit `any` types.

src/feature/common/Sidebar/SidebarNode.tsx
```diff
@@ -86,12 +89,19 @@ const SidebarNode = ({
   autogeneratedCategory,
   icon,
 }: SidebarNodeProps) => {
-  if (node.name === "ROOT") {
+  debugger;
```
JeffDotPng, line 92: Should be removed

JeffDotPng (commented):
Very nice, just a couple small things

src/utils/ManifestLoader.ts
```diff
@@ -107,3 +108,29 @@ try {
 }
 
 export { nodeSection };
+
+export function isLeaf(obj: any): obj is Leaf {
```
JeffDotPng, line 112: Not a big fan of `any`, since it requires you to disable ESLint. Since it only makes sense to call these functions with tree nodes, I think you could define a `type TreeNode = RootNode | RootChild | ParentNode | Leaf` then change the type signature to that. For example in `isLeaf`, it would just become
```ts
export function isLeaf(obj: TreeNode): obj is Leaf {
  return "name" in obj && "key" in obj && "type" in obj && !obj.children;
}
```

src/utils/ManifestLoader.ts
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable */
```
JeffDotPng, line 1: Should remove once for types for type guard functions are fixed

itsjoeoui (approved):
lgtm

PR #856, opened 2023-08-31
Shift nodes manifest generation functionality to backend

smahmed776:


captain/utils/nodes_metadata.py
```diff
@@ -0,0 +1,46 @@
+import fnmatch
+import json
+import os
+
+nodes_dir = "PYTHON/nodes"
+
+# The pattern to match for Python files
+pattern = "*.py"
+
+badbadnotgood = ["VCTR.py", "__init__.py", ".DS_Store"]
```
itsjoeoui, line 10: why is VCTR.py in this list, seems like it is not something "harmful"?

itsjoeoui (dismissed):

captain/utils/nodes_metadata.py
smahmed776: Actually, we can remove that.

JeffDotPng (approved):
lgtm

PR #857, opened 2023-09-01
Make node delay use milliseconds instead of seconds

JeffDotPng:
Previously it was impossible to set the node delay as a fraction of a second, so this addresses the issue by just using milliseconds (input is still locked to integer values)


itsjoeoui (approved):
lgtm

PR #860, opened 2023-09-04
Add max workers setting

izi-on:
Temporary band-aid on a bigger issue: refer to "STU-122: LOOP +1 and LOOP INDEX mismatch" on Linear

This PR will default single-threading for the app (1 worker) but can be adjusted. With the default, we won't run into concurrency issues and race conditions.


itsjoeoui (approved):
lgtm

PR #863, opened 2023-09-06
Allow returning no plotly figure from job result

JeffDotPng:
For huge performance improvement related to: https://github.com/flojoy-ai/python/pull/99


izi-on (approved):
lgtm

PR #875, opened 2023-09-12
[WIP] Get pyinstaller working

smahmed776:


smahmed776:
Redundant now as we decided to ship micromamba executable with electron package and initialize backend with nodejs child process, see #894

PR #886, opened 2023-09-20
Topology test

smahmed776:


captain/tests/topology_test.py
```diff
@@ -0,0 +1,45 @@
+import unittest
+from queue import Queue
+from .test_apps.sample_app import graph as sample_app_graph
+from captain.models.topology import Topology
+from captain.types.worker import JobInfo
+from flojoy import JobSuccess
+
+
+class TopologyTest(unittest.TestCase):
```
JeffDotPng, line 9: could you add a test that the end result is what was expected? i.e use numpy to compute the same values that the topology is computing and assert that they are equal

captain/tests/topology_test.py
```diff
@@ -0,0 +1,102 @@
...
+        waiting_time = time.time()
+        while topology.is_finished() != True:
+            if (time.time() - waiting_time) < self.MAX_TIMEOUT:
+                time.sleep(0.5)
+            else:
+                cleanup()
+                raise TimeoutError(
+                    f"Flowchart failed to run within {self.MAX_TIMEOUT} seconds!"
+                )
+        cleanup()
```
JeffDotPng, line 94: This seems very convoluted just to test functionality like running an app.. probably a sign the codebase is not very testable, definitely needs some refactoring in the future (probably fine for now though)

It seems we don't have an easy way to inspect the results of nodes? Because it would be good to assert that the result of each node is what it's supposed to be..

JeffDotPng (approved):

captain/tests/topology_test.py
smahmed776: @JeffDotPng I totally agree that we should probably refactor the code so it's easier to test. But here's the thing: we've only been getting results for Plotly() and TextBlob() lately and ignoring all other `DataContainer` types, so, at least for now, trying to check the node results isn't gonna work.

JeffDotPng: The actual problem isn't that, the "result" from `JobSuccess` was always a plotly figure/text blob. The actual data from the job is only written to the `JobService`, which just writes to `Dao` which is just global state and thus very hard to test.

smahmed776: > The actual problem isn't that, the "result" from `JobSuccess` was always a plotly figure/text blob. The actual data from the job is only written to the `JobService`, which just writes to `Dao` which is just global state and thus very hard to test.

@JeffDotPng  You're right, At the moment, I believe we're good with the current testing setup. We can tackle the mass refactoring down the road.

PR #910, opened 2023-10-10
npm -> pnpm

itsjoeoui:
Also improved caching for CI, now we can install python + node deps in under 15 seconds with cache, pretty great.


smahmed776 (approved):
LGTM

PR #912, opened 2023-10-11
Add webcam support for mac

TheBigSasha:
Add support for MacOS webcams to the new Hardware Devices view.
Screenshot 2023-10-11 at 12 28 59 AM

NOTE:
_Needs testing on Windows and Linux in case the installation of pyobjc breaks things there_

### DIFF
--- a/captain/services/hardware.py
+++ b/captain/services/hardware.py
@@ -3,12 +3,12 @@ import pyvisa
 import cv2
 import subprocess
 from sys import platform
+if platform in ["darwin"]:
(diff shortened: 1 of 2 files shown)


TheBigSasha:
Builds in non Mac systems fail, try dynamically installing the pyobjc library?
``import pip

def install(package):
    if hasattr(pip, 'main'):
        pip.main(['install', package])
    else:
        pip._internal.main(['install', package])

# Example
if __name__ == '__main__':
    install('argh') ``

PR #923, opened 2023-10-19
Don't run studio if there is already a server running

itsjoeoui:
- add a check to see if there is already existing backend running
- update deps and shell script
- add a port check
- Properly relaunch


smahmed776 (approved):
LGTM

PR #953, opened 2023-11-16
Auto sync frontend blocks with manifest

JeffDotPng:


captain/services/consumer/blocks_watcher.py
```diff
@@ -0,0 +1,18 @@
...
+
+class BlocksWatcher:
+    def __init__(self) -> None:
+        self.ws = ConnectionManager.get_instance()
+
+    async def run(self):
+        blocks_path = get_blocks_path()
+        logger.info(f"Starting file watcher for blocks dir {blocks_path}")
+
+        async for change in awatch(blocks_path):
```
itsjoeoui, line 15: one concern is that, sometimes `__pycache__`, or maybe even more other stuff, gets created in this directory, which may also gets picked up by this watcher, causing unneeded updates. I think we should only watch those important files (kinda like having a whitelist of files that we should watch)

itsjoeoui (changes requested):

PR #972, opened 2023-11-30
joey/blocks version

itsjoeoui:
- feat: download the same blocks version as studio
- feat: make sure to always grab the latest tag

### DIFF
--- a/src/main/blocks.ts
+++ b/src/main/blocks.ts
@@ -48,8 +48,6 @@ const getBlocksVersion = (): string => {
 };
 
 export const isBlocksOutdated = (): boolean => {
-  console.log(version);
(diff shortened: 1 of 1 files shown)


src/main/blocks.ts
```diff
@@ -5,6 +5,7 @@ import axios from "axios";
 import AdmZip from "adm-zip";
 import { sendToStatusBar } from "./logging";
 import { ncp as cpFolder } from "ncp";
+import { version } from "../../package.json";
```
smahmed776, line 8: I think using `app.getVersion()` would be a good idea than manually importing package.json file

smahmed776 (approved):
LGTM

PR #1007, opened 2023-12-11
Replace use-file-picker with electron API

smahmed776:


package.json
itsjoeoui, line 1: Do you know why lots of dependencies got downgraded here?

itsjoeoui: Gotcha! Let me fix this later today, I think I know what is wrong.

smahmed776: Sounds good

JeffDotPng (approved):

PR #1013, opened 2023-12-13
Simple Block Code Editor

itsjoeoui:
CleanShot 2023-12-12 at 20 04 09@2x
CleanShot 2023-12-12 at 20 04 20@2x
CleanShot 2023-12-12 at 20 04 46@2x


src/main/utils.ts
```diff
@@ -133,3 +136,34 @@ export const openFilePicker = (): Promise<
...
+  return fs.readFileSync(filePath, { encoding: "utf-8" }).toString();
+};
+
+export const saveFileToFullPath = async (
+  filePath: string,
+  content: string,
+): Promise<boolean> => {
+  fs.writeFileSync(filePath, content);
+  return true;
+};
```
JeffDotPng, line 169: why return boolean if it's always going to be true?

src/renderer/routes/flow_chart/views/BlockModal.tsx
```diff
@@ -117,6 +120,15 @@ const BlockModal = ({
...
           Python code
         </h2>
+        <Button
+          onClick={async () => {
+            await window.api.openEditorWindow(blockFullPath);
+          }}
+          data-testid="btn-edit-python"
+        >
+          Edit Python Code
+        </Button>
```
JeffDotPng, line 130: image
this button could afford to be a bit less wide

src/main/utils.ts
itsjoeoui: Fixed! Switched to the `Result` type instead.

src/renderer/routes/flow_chart/views/BlockModal.tsx
itsjoeoui: Done.

src/main/utils.ts
```diff
@@ -163,7 +164,7 @@ export const loadFileFromFullPath = async (
 export const saveFileToFullPath = async (
   filePath: string,
   content: string,
-): Promise<boolean> => {
+): Promise<Result<void>> => {
   fs.writeFileSync(filePath, content);
-  return true;
+  return { ok: true, data: undefined };
```
JeffDotPng, line 169: I would add functions (constructors) to make this type so you don't have to write this `ok: true` explicitly
e.g 
```typescript
function Ok<T>(data: T): Result<T, E> {
    return { ok: true, data }
}
function Err<E>(error: E): Result<T, E> {
    return { ok: false, error }
}
```

src/main/utils.ts
```diff
@@ -133,3 +137,34 @@ export const openFilePicker = (): Promise<
...
+  filePath: string,
+): Promise<string> => {
+  return fs.readFileSync(filePath, { encoding: "utf-8" }).toString();
+};
+
+export const saveFileToFullPath = async (
+  filePath: string,
+  content: string,
+): Promise<Result<void>> => {
+  fs.writeFileSync(filePath, content);
```
JeffDotPng, line 168: you need to try/catch this as well, the only result your function is ever going to return will be Ok(()) but it can still throw an error as usual
```typescript
try {
  fs.writeFileSync(filePath, content);
  return Ok()
}
catch (e) {
  return Err(e);
}

JeffDotPng (approved):

PR #1019, opened 2023-12-15
Mixpanel

itsjoeoui:


src/renderer/services/MixpanelServices.ts
```diff
@@ -35,31 +38,36 @@ export const sendProgramToMix = (
...
+  dataType: string = "data",
 ) => {
+  if (await isCI()) return;
   if (enable) {
     try {
-      mixpanel.track(event, { [dataType]: data });
+      mixpanel.track(
+        event,
+        typeof data === "string" ? { [dataType]: data } : data,
+      );
```
JeffDotPng, line 57: I feel like you could just only use `Record<string, unknown>` then make the caller specify the key instead of passing it as a separate parameter and converting it into a record in the function.

src/main/store.ts
```diff
@@ -2,11 +2,13 @@ import Store from "electron-store";
 
 type TypedStore = {
   poetryOptionalGroups: string[];
+  setupStarted: number;
 };
 
 export const store = new Store<TypedStore>({
   defaults: {
     poetryOptionalGroups: [],
+    setupStarted: 0,
```
JeffDotPng, line 11: Should this be in store? it doesn't ever need to be persisted, and every time you use it you need to set it to 0 again after, I feel like it could just be a value you attach to the `global` object instead at startup, what do you think?

smahmed776: You're right, attaching to global object makes more sense. Thanks

src/renderer/services/MixpanelServices.ts
smahmed776: I was initially considering backward compatibility to avoid altering existing implementations of this function in other places. However, I'll now proceed and implement this, as you suggested. Thanks!

JeffDotPng (approved):

PR #1030, opened 2023-12-26
Fix an issue when there is a space in python path

smahmed776:


src/main/python/index.ts
```diff
@@ -83,7 +84,16 @@ export async function pipxEnsurepath(): Promise<void> {
 
 export async function installPoetry(): Promise<void> {
   const py = process.env.PY_INTERPRETER ?? "python";
-  await execCommand(new Command(`${py} -m pipx install poetry --force`));
+  const localDir = `${os.homedir()}/.local`;
```
itsjoeoui, line 87: I think it is better to replace with `path.join(os.homedir(), '.local');`?

itsjoeoui (changes requested):

src/main/python/index.ts
smahmed776: Yeah, sounds good. Modified as requested. Thanks

JeffDotPng (approved):

PR #1034, opened 2024-01-08
New cloud client

39bytes:


pkgs/flojoy/flojoy/flojoy_cloud.py
```diff
@@ -1,475 +1,344 @@
...
+        self,
+        method: HttpMethod,
+        endpoint: str,
+        *,
+        query_params: dict[str, int | str] | None = None,
+        data: str | None = None,
+        body: dict | None = None,
+        parse_as: type[T] | TypeAdapter[U] | None = None,
+    ) -> T | U | None:
+        res = requests.request(
```
itsjoeoui, line 185: Should probably use httpx instead of requests imo. 
https://www.python-httpx.org/

pkgs/flojoy/flojoy/flojoy_cloud.py
```diff
@@ -1,475 +1,344 @@
...
+            )
+
+        if parse_as is None:
+            return None
+
+        return self.__parse(parse_as, res.text)
+
+    """Test Endpoints"""
+
+    def create_test(
```
itsjoeoui, line 207: https://www.python-httpx.org/advanced/#sharing-configuration-across-requests
With httpx you can share client like this

itsjoeoui (changes requested):

pkgs/flojoy/flojoy/flojoy_cloud.py
```diff
@@ -1,475 +1,344 @@
...
+
+MeasurementData = Boolean | Dataframe
+
+
+T = TypeVar("T", bound=BaseModel)
+U = TypeVar("U")
+
+
+class CloudModel(BaseModel):
+    model_config = ConfigDict(alias_generator=to_camel)
```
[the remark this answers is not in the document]
JeffDotPng: Cloud server uses camelCase for fields because it's in javascript, python client uses snake_case because python, this just does the name conversions automatically

itsjoeoui (approved):
lgtm

PR #1046, opened 2024-01-16
[New Block] Export DataFrame, Matrix, and Pairs to a .mat file

LatentDream:
@dstrande, can you run the app.json and check that the file generated from the app can be loaded into Matlab, I don't have it on my computer.


blocks/DATA/EXPORT/EXPORT_MATLAB/EXPORT_MATLAB.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Optional
+from scipy.io import savemat
+from typing import Literal
+
+
+@flojoy
+def EXPORT_MATLAB(
+    dc: OrderedPair | OrderedTriple | DataFrame | Matrix,
+    dir: Directory,
+    filename: str = "exported.csv",
```
dstrande, line 21:
```suggestion
    filename: str = "exported.mat",
```

dstrande (changes requested):

dstrande (approved):

PR #1051, opened 2024-01-22
STU-219 | NI-CompactDAQ - read Analog

LatentDream:
**Currently supported:**
  -  Read a number of samples from analog input channel(s)
  -  Read from an input stream, and re-use the Buffer    

> So a program with a NI DAQ involves the following steps:

    1. Create a Task and Virtual Channels
    2. Configure the Timing Parameters
    3. Start the Task
    4. Perform a Read operation from the DAQ
    5. Perform a Write operation to the DAQ _not yet implemented_
    6. Stop and Clear the Task.


LatentDream:
https://github.com/flojoy-ai/studio/assets/56351875/eef8c217-35f9-4415-81a3-702830798153

Reading from stream compact DAQ example

LatentDream:
The more I think about it, the more it makes sense to split `CREATE_TASK_CHANNEL` into `CREATE_TASK` and `ATTACH_CHANNEL`. This way, a user can assign a name to a task, and he/she will need to input the same name in blocks like attaching a channel or reading from all channels attached to a task.

The current implementation works fine for a single module, but dealing with a multi-chassis module and synchronizing those modules together will become more challenging with the existing implementation.

**Putting it back in Draft, I'll do the change tomorrow**

LatentDream:
@dstrande, let me know if there is more thing todo. 
### List
- [] Toggle switch in the hardware tab for NI-Devices **Off by default**
- [] UI to display NI-Device in the hardware tab

dstrande:
> @dstrande, let me know if there is more things todo.
> ### List
> 
>     * [ ]  Toggle switch in the hardware tab for NI-Devices **Off by default**
> 
>     * [ ]  UI to display NI-Device in the hardware tab

Thanks! looks right

LatentDream:
2024-01-29 21_33_49-Flojoy Studio
2024-01-29 21_33_30-Flojoy Studio


Work in progress, missing:
- [ ] Backend implementation
- [ ] Changing devices name and id to be different
- [ ] Message here and there on how to activate the driver dependent devices discovery

LatentDream:
# Demo
https://github.com/flojoy-ai/studio/assets/56351875/d665cf52-4fb7-4152-afc9-1583c52de816

@dstrande, @39bytes, After the changes, let me know if you see something off

dstrande (approved):
Awesome LGTM!

PR #1054, opened 2024-01-24
CAN-BUS Support

LatentDream:
## Hardware
- [PEAK-USB](https://www.peak-system.com/PCAN-USB.199.0.html?&L=1)
    - [X] Tested on Windows with [PEAK-Can Drivers](https://www.peak-system.com/Drivers.523.0.html?&L=1) (same driver on linux)
    - [X] Tested on Mac with [Mac-CAN driver](https://mac-can.github.io/drivers/)

 - [CANABLE](https://canable.io/)
    - [X] Change old block to be compatible with new CAN Utils 

## Blocks Added
- [X] Discover to a CAN Bus devices using a PCAN interface
- [X] Connect to a CAN Bus using a PCAN interface
- [X] Read from CAN Bus
- [X] Create can message
- [X] Add Kernel level Filter for CAN Bus Device
- [X] Remove kernel filter
- [X] Send Periodic message
- [X] Stop Periodic message
- [X] Filter message (ID or ERROR) (for log file)
- [X] Load a CAN Database file (DBC)
- [X] Decode a CAN Message using a DBC
- [X] Write to [ASAM MDF file](https://www.asam.net/standards/detail/mdf/wiki/)
- [X] Read ASAM MDF file

## Demo
- [X] Read from PEAK usb and log messages to a MF4 log file with a filter apply at the kernel level
- [X] Send message to a CANABLE device, single message and periodic messages

## Block Modified
- The Append function was accepting a `DataFrame`, but there was no implementation.

# Core App Modification
- Enabled the use of `HardwareDevice` and `HardwareConnection` as a general way to save connections.
       - I know they are supposed to be abstract classes, but I felt it was appropriate instead of adding a new class each time for this simple behavior. I can revert this change and create actual `CustomHardwareDevice` and `CustomHardwareConnection` if needed.


LatentDream:
**In Draft Because**: Waiting on some boards to test the whole thing with real data. Currently, it has only been tested with virtual data. Right now: The PEAK-USB connection is successful but times out because it's not connected to anything.

LatentDream:
# Example
Peek connected to a real CAN bus system and read all the data on the bus. 

> The errors in the message are normal; they are present because there is no thermocouple on the sensor input. Therefore, the sensor is informing the receiver that those inputs aren't good.

image

blocks/HARDWARE/PROTOCOLS/CAN/UTILS/DECODE/DECODE.py
```diff
@@ -0,0 +1,40 @@
+from flojoy import flojoy, Stateful, DataFrame
+import pandas as pd
+import logging
+
+
+@flojoy(deps={"python-can": "4.3.1", "cantools": "39.4.2"})
+def DECODE(
```
dstrande, line 7: I'd change this to DECODE_CAN or something because we will probably have a few other blocks, like DECODE_I2C.

dstrande (changes requested):
LGTM other than that one change

LatentDream:
@dstrande I've merged the [STU-249 CAN-Utils](https://linear.app/flojoy/issue/STU-249/[block]-can-utils) with this PR since I've updated most of the CAN block names. This PR was starting to lose its relevance.

dstrande:
> Enabled the use of HardwareDevice and HardwareConnection as a general way to save connections. - I know they are supposed to be abstract classes, but I felt it was appropriate instead of adding a new class each time for this simple behavior. I can revert this change and create actual CustomHardwareDevice and CustomHardwareConnection if needed.

Where is this change?

LatentDream:
> > Enabled the use of HardwareDevice and HardwareConnection as a general way to save connections. - I know they are supposed to be abstract classes, but I felt it was appropriate instead of adding a new class each time for this simple behavior. I can revert this change and create actual CustomHardwareDevice and CustomHardwareConnection if needed.
> 
> Where is this change?

It got merge by another PR that needed the same behavior 

https://github.com/flojoy-ai/studio/pull/1051/files#diff-571ede2e8b83f5840515d1787c6926c4a905b8d37e428ab4eee3d77fda30fdf2R55

dstrande:
I think I'll take a good, long, look at this tomorrow morning if that's OK with you

LatentDream:
> I think I'll take a good, long, look at this tomorrow morning if that's OK with you

Yeah that's perfect, No rush in merging. and thanks for the real review :)

blocks/HARDWARE/PROTOCOLS/CAN/BUS/PEAK_DETECT_AVAILABLE_DEVICES/PEAK_DETECT_AVAILABLE_DEVICES.py
```diff
@@ -0,0 +1,38 @@
...
+
+    This blocks is use to detect available PCAN devices and returns a string containing the device name and the channel number.
+    → Use the channel number to configure to connect to PEAK can hardware and other devices that support the PCAN interface using the PCAN_CONNECT block.
+
+    For Windows and Linux compatibilities, the following Drviers are required:
+        → https://www.peak-system.com/Drivers.523.0.html?&L=1
+    For MacOS compatibilities, the following Drivers are required:
+        → https://mac-can.github.io/
+
+    Parameters
```
dstrande, line 18: I believe when params are none, we just remove this section.

blocks/HARDWARE/PROTOCOLS/CAN/BUS/RECEIVE_CAN_MESSAGE/RECEIVE_CAN_MESSAGE.py
```diff
@@ -0,0 +1,44 @@
+from flojoy import flojoy, DataContainer, DeviceConnectionManager, Stateful
+from typing import Optional
+import can
+
+
+@flojoy(deps={"python-can": "4.3.1"})
+def RECEIVE_CAN_MESSAGE(
+    CAN_address: str,
```
dstrande, line 8: We should probably include these in the front end at some point

blocks/HARDWARE/PROTOCOLS/CAN/BUS/REMOVE_CAN_BUS_FILTER/REMOVE_CAN_BUS_FILTER.py
```diff
@@ -0,0 +1,30 @@
...
+    CAN_address: str, default: Optional[DataContainer] = None
+) -> Optional[DataContainer]:
+    """Remove all filters attached to a CAN bus connection.
+
+    A connection to the device is required. Use a CAN_CONNECT block to connect to a CAN device.
+
+    Parameters
+    ----------
+    CAN_address : str
+        The CAN bus address to attach the filter to.
```
dstrande, line 17: Are you attaching a filter or removing one?

blocks/HARDWARE/PROTOCOLS/CAN/BUS/SET_CAN_BUS_FILTER/SET_CAN_BUS_FILTER.py
```diff
@@ -0,0 +1,45 @@
...
+    A connection to the device is required. Use a CAN_CONNECT block to connect to a CAN device.
+
+    Parameters
+    ----------
+    CAN_address : str
+        The CAN bus address to attach the filter to.
+    can_id : int
+        The ID of the message to filter.
+    can_mask : int
+        todo
```
dstrande, line 28: TODO

blocks/HARDWARE/PROTOCOLS/CAN/BUS/STOP_SEND_PERIODIC_CAN_MESSAGE/STOP_SEND_PERIODIC_CAN_MESSAGE.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+from flojoy import flojoy, DeviceConnectionManager, DataContainer
+import can
+
+
+@flojoy(deps={"python-can": "4.3.1"})
+def STOP_SEND_PERIODIC_CAN_MESSAGE(
```
dstrande, line 7: Reference this node in the docstring for SEND_PERIODIC and visa versa?

blocks/HARDWARE/PROTOCOLS/CAN/UTILS/DECODE_CAN_MESSAGE/DECODE_CAN_MESSAGE.py
```diff
@@ -0,0 +1,46 @@
...
+        The database to use for decoding the message.
+    messages : Stateful
+        The message to DECODE_CAN. Must be a can.Message object.
+    ignore_undefined_id_error : bool
+        If True, ignore undefined id error. Default is False.
+
+    Returns
+    -------
+    DataFrame : DataFrame
+        Return dataframe containing the DECODE_CANd message.
```
dstrande, line 28: typo?

blocks/HARDWARE/PROTOCOLS/CAN/UTILS/MF4_WRITE/MF4_WRITE.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import Optional
+import can
+
+
+@flojoy(deps={"python-can": "4.3.1", "asammdf": "7.4.1"})
+def MF4_WRITE(
+    MF4_writer: Stateful,
+    messages: Stateful,
+) -> Optional[DataContainer]:
+    """Create a writer for the MF4 format.
```
dstrande, line 11: Same as CREATE WRITER?

dstrande (changes requested):
A couple typos and suggestions

blocks/HARDWARE/PROTOCOLS/CAN/BUS/RECEIVE_CAN_MESSAGE/RECEIVE_CAN_MESSAGE.py
LatentDream: I agree, but there are various CAN interfaces, some requiring specific strings like for PEAK to instanciate the connection with the driver, and others just requiring the serial address like CANABLE. Since we don't provide connections for all possible interfaces, it could cause extensibility problem if someone needs an interface that is not currently supported and try to add the CONNECT block to its interface and try to use the other CAN blocks.

blocks/HARDWARE/PROTOCOLS/CAN/UTILS/DECODE_CAN_MESSAGE/DECODE_CAN_MESSAGE.py
LatentDream: Ahahah, yes, when I switched from DECODE to DECODE_CAN, I forgot to include the case-sensitive option

blocks/HARDWARE/PROTOCOLS/CAN/BUS/PEAK_DETECT_AVAILABLE_DEVICES/PEAK_DETECT_AVAILABLE_DEVICES.py
LatentDream: Just tried it out; the docs generator requires it :)

dstrande (approved):

LatentDream:
I forgot to squash everything when merging :/

PR #1071, opened 2024-02-07
STU-251 | New cloud block integration

LatentDream:
- [X] Converted Upload Block 
    - Support type that can be casted to a Dataframe since only boolean and dataframe are supported for new
    - Will need Image, scalar, float support. To be added when cloud support it.
- [X] Converted Download Block
     - Only return Dataframe or Boolean. 
    - _Should I add a check on the column name and try to cast them in OrderedPair, OrederedTripple ... ? Or I wait and we add those type on the cloud side ?_

- [X] Removed Cloud from flojoy package to use the flojoy_cloud package instead


itsjoeoui (approved):
lgtm

PR #1082, opened 2024-02-14
Environment variables as block input

LatentDream:
# New Input Type
### Those variables:
Screenshot 2024-02-13 at 8 06 41 PM

### Can appear in the Block input panel
Screenshot 2024-02-13 at 8 06 25 PM

# Other
- Using get_env_var to retreive the value was causing circular dependencies, so I splited them, it might exists a better solution


LatentDream:
TS code is shitty, I need to spend some time on it tomorrow. Just making sure CI is good before.
I need to make some test also before converting to "Ready for Review"

pkgs/flojoy/flojoy/env_var.py
```diff
@@ -0,0 +1,72 @@
...
+    if key not in keys:
+        return
+
+    keys.remove(key)
+
+    with open(file_path, "w") as f:
+        f.write(",".join(keys))
+
+    kr = get_keyring()
+    kr.delete_password("flojoy", key)
```
39bytes, line 72: Since we're only ever reading the variables from the system keyring, do we even still need to write it to the credentials file? I don't think anything actually reads from that file :thinking:

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    const fetchCredentials = async () => {
+      try {
+        const res = await baseClient.get("env");
+        const keys = res.data.map((d) => d.key);
+        setSecrets(keys);
+        setFound(keys.length > 0);
+        if (found) {
+          value ??= keys[0];
+        }
```
39bytes, line 30: No need to do this, to actually change some state you need to use the setter and not assignment

pkgs/flojoy/flojoy/env_var.py
LatentDream: They are use to retrieve the variable in get .../env/ route. Not sure with the route don't read it directly from keyring ? :/
I can remove it and just use Keyring

LatentDream: Oh, it's so the app can know all the env vars the user has set. Their is no "get_all_passwords" from Keyright. One need to know the name of the key to retrieve. But the current implementation is not the most secure

smahmed776: @39bytes @LatentDream The purpose of writing to the credential file was to store the environment keys set by the user. This enables us to display them on the Environment modal, ensuring users are aware if a secret for this environment has already been set. I think we can keep it as keyring doesn't allow to list all keys...

LatentDream: Agreed:)

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState, useEffect } from "react";
+
+export type SelectProps = {
+  onValueChange: (value: string) => void;
+  value: string | number | boolean | null | undefined;
+};
+
+export const SecretSelect = ({ onValueChange, value }: SelectProps) => {
+  const [secrets, setSecrets] = useState<string[]>([]);
+  const [found, setFound] = useState<boolean>(false);
```
smahmed776, line 19:
```suggestion

```

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  useEffect(() => {
+    const fetchCredentials = async () => {
+      try {
+        const res = await baseClient.get("env");
+        const keys = res.data.map((d) => d.key);
+        setSecrets(keys);
+        setFound(keys.length > 0);
```
smahmed776, line 27:
```suggestion

```

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }
+    };
+    fetchCredentials();
+  }, []);
+
+  return (
+    <Select onValueChange={onValueChange}>
+      <SelectTrigger
+        className="border-none bg-background focus:ring-accent1 focus:ring-offset-1 focus-visible:ring-accent1 focus-visible:ring-offset-1 "
+        disabled={!found}
```
smahmed776, line 39:
```suggestion
        disabled={!secrets.length}
```

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+      >
+        <SelectValue
+          placeholder={found ? value : "No environment variable found"}
```
smahmed776, line 42:
```suggestion
          placeholder={secrets.length ? value : "No environment variable found"}
```

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+        />
+      </SelectTrigger>
+      <SelectContent className="max-h-72">
+        <SelectGroup>
+          {found &&
```
smahmed776, line 47:
```suggestion
          {!!secrets.length &&
```

smahmed776 (changes requested):

LatentDream:
> I believe we should avoid sending secret values in the frontend, as sensitive information like AWS credentials could be exposed to other users. Instead, we can transmit keys and display them with masked characters (e.g., *****) in the Environment modal. Alternatively, we can implement authentication on the /env/ route to prevent Viewers from accessing secrets.
> 
> We can just return keys array from [here](https://github.com/flojoy-ai/studio/pull/1082/files#diff-71b8c9e2d908a10e60e03e0ff08743355d1abddc0ade6783d909d3526fd7a426R315) without retrieving values from keyring.
> 
> cc @itsjoeoui

I took the same route as the _Set Environment variable modal_ uses. The "See the env var" feature need this, but we could just provide the key instead with get /env/. And do another request get /env/{key} when the user want to see it from the Env. Var. Modal.

I also realize that a "Viewer" user has access to the "View env var" feature:
Screenshot 2024-02-15 at 11 26 34 AM

itsjoeoui:
Hey @smahmed776 @LatentDream I think it will be okay to view it as long as that is only for admins, and not viewers!

src/renderer/routes/flow_chart/components/node-edit-menu/SecretSelect.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [secrets, setSecrets] = useState<string[]>([]);
+
+  useEffect(() => {
+    const fetchCredentials = async () => {
+      try {
+        const res = await baseClient.get("env");
+        const keys = res.data.map((d) => d.key);
+        setSecrets(keys);
+      } catch (error) {
+        console.log(error);
```
itsjoeoui, line 27:
```suggestion
        console.error(error);
```

itsjoeoui (changes requested):
lgtm

smahmed776 (approved):

itsjoeoui:
> > Hey @smahmed776 @LatentDream I think it will be okay to view it as long as that is only for admins, and not viewers!
> 
> @itsjoeoui Limiting access in the frontend to only admins seeing the secrets is effective. However, APIs can still be accessed externally if we don't enforce restrictions in the backend too. We can work on that in a separate PR obviously.

Yeep, I think we shall limit the access on the frontend for now, one thing we probably want to do is to integrate with HashiCorp Vault for more secure env management in an enterprise settings

itsjoeoui (approved):
lgtm

PR #1083, opened 2024-02-14
abstract test sequencer run & raise error when accessing identifier before eval

izi-on:


captain/parser/bool_parser/expressions/exceptions.py
```diff
@@ -44,3 +44,9 @@ class InvalidIdentifier(Exception):
     def __init__(self, message=""):
         self.message = "Invalid identifier: " + message
         super().__init__(self.message)
+
+
+class EarlyIdentifier(Exception):
+    def __init__(self, message=""):
+        self.message = f"Cannot access identifier {message} before evaluation"
```
LatentDream, line 51: The error is good, but a little bit too "comp sci" for a user. Maybe something like:
```
f"Unable to access test result for {message} as it's executed afterward"
```

captain/utils/test_sequencer/run_test_sequence.py
```diff
@@ -101,56 +108,96 @@ async def _stream_result_to_frontend(
...
+                {
+                    "if": (None, _case_if_node),
+                },
+            ),
+        },
+    )
+    matcher, cur = map_to_handler
+    while not callable(cur):
+        matcher, cur = cur[node.__dict__[matcher]]
+    children_getter, test_result = await cur(node)  # sus name for the variable
```
LatentDream, line 175: Ahahha, would remove the comment tho

LatentDream (changes requested):
lgtm, just some small things.
Also not sure we need a __init__ in each directory ?

itsjoeoui (approved):
lgtm

PR #1091, opened 2024-02-20
Pi performance

LatentDream:
The motion-reduce feature is not easy to turn on in PI, so I included some logic to remove the blur on Linux Arm by using the exposed API from Chromium.

We can go this route or it's possible to query the OS to read the hardware model from `/proc/cpuinfo` from the main Electron process. This way, we can apply it only to the Pi model ? Should I switch for that ?


izi-on (dismissed):
lgtm, although I can't rlly test this since I don't have a Pi. Would rather have someone else review this too

PR #1095, opened 2024-02-21
Fix: Run Test sequencer without test | Init tree as a root with children 

LatentDream:
Invalid pydantic modal was crashing the socket


smahmed776 (approved):
looks good

PR #1097, opened 2024-02-21
Show Python code for test in a modal

smahmed776:


src/renderer/hooks/useTestImport.ts
```diff
@@ -5,10 +5,14 @@ import { map } from "lodash";
...
 
-function parseDiscoverContainer(data: TestDiscoverContainer) {
+function parseDiscoverContainer(
+  data: TestDiscoverContainer,
+  sourceCode: string,
+) {
   return map(data.response, (container) => {
     const new_elem: Test = {
       ...container,
+      sourceCode,
```
izi-on, line 15: how will we handle this when the user wants to change the python code in the file? 

Suggestions:
- use the reactive paradigm: detect or continuously monitor for changes in the file and define a callback function to populate the "sourceCode" field for an element.

smahmed776: @izi-on Very good question. I'd go with fetching code content each time modal is open as this seems to be simplest yet effective..

izi-on (changes requested):

src/renderer/hooks/useTestImport.ts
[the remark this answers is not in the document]
smahmed776: Good point 👍 It makes sense as we have access to test path and fetching source code every time opening the modal so there's no need to include that in testElement. Thanks!

LatentDream (approved):

PR #1104, opened 2024-02-22
Limit access to secrets to the admins only

smahmed776:


src/renderer/lib/base-client.ts
```diff
@@ -1,10 +1,12 @@
...
 import axios from "axios";
 
-export const baseClient = axios.create({
-  baseURL: API_URI,
-  withCredentials: true,
-  headers: {
-    "Content-Type": "application/json",
-  },
-});
+export const baseClient = (basicAuth?: `${string}:${string}`) =>
```
itsjoeoui, line 4: when you do `withCredentials: true`, credentials such as cookies, auth headers, etc will already be included with the request, assume the cookie is configured properly you do not need to attach it again like this. Now also every time you are initializing a new client upon request which is not efficient

itsjoeoui (changes requested):
lgtm

src/renderer/lib/base-client.ts
smahmed776: Thanks, I've adopted cookie-based auth system for backend.

itsjoeoui (approved):
lgtm

PR #1111, opened 2024-02-25
STU-282 & STU-277 & STU-281 & STU-284 & STU-271 | Test Sequencer

LatentDream:
## FYI
The way Pytest Work, I can only retreive one dependency at the time. And I'm greping the return Error, so a missing "relative import" could be confuse with a library that could be install.

---

## This PR is about
- A way to import library for Test Sequencer
- Warnings when no Pytest test is found
- Running Python Script
- Mouse hover `FAIL` test to show the what wront
- Cancel Test Sequencer Run


src/renderer/routes/flow_chart/views/DepManagerModal.tsx
```diff
@@ -31,40 +33,74 @@ const DepManagerModal = ({
...
 
   const handleUpdate = useCallback(async () => {
     setIsFetching(true);
+    setMsg("Fetching dependencies...");
     const deps = await window.api.poetryShowTopLevel();
+    const userDeps = await window.api.poetryShowUserGroup();
     const groups = await window.api.poetryGetGroupInfo();
     setAllDependencies(deps);
+    setUserDependencies(userDeps);
+    console.log(userDependencies);
```
LatentDream, line 55:
```suggestion
```

captain/internal/manager.py
```diff
@@ -19,7 +21,25 @@ def __init__(self):
 
 # Manager for Test Sequencer activities
 class TSManager(WSManager):
-    pass  # could potentially add more things in the future for this, that's why we keep it as such
+
+    def __init__(self):
+        self.runner: asyncio.Runner | None = None  # holds the running sequencer
```
izi-on, line 26: it's better that the "runner" is not a asyncio runner by rather a thread. The reason being if we want to cancel a long test atom, it will wait for it to finish first. (Do we leave it like this for now?)

captain/utils/pytest/discover_tests.py
```diff
@@ -10,14 +10,37 @@
...
 
 
-def discover_pytest_file(path: str, one_file: bool, return_val: list):
+def check_missing_imports(report: JSONReport):
+    missing_lib = set()
+    if not report.report:
+        return missing_lib
+    if "collectors" not in report.report:
+        return missing_lib
+    for element in report.report["collectors"]:
```
izi-on, line 22: best not to use as a dict, but rather access like a class object (for developer experience)

captain/utils/test_sequencer/run_test_sequence.py
```diff
@@ -104,6 +149,7 @@ async def _stream_result_to_frontend(
     is_saved_to_cloud: bool = False,
     error: str | None = None,
 ):
+    logger.error(f"streaming result to frontend {state} {test_id} {result} {time_taken} {is_saved_to_cloud} {error}")
```
izi-on, line 152: why is this .error() ?

captain/utils/test_sequencer/run_test_sequence.py
```diff
@@ -138,15 +184,20 @@ async def _case_root(node: TestRootNode, **kwargs) -> Extract:
 async def _case_test(node: TestNode, **kwargs) -> Extract:
     # TODO: support run in parallel feature
     await _stream_result_to_frontend(MsgState.RUNNING, test_id=node.id)
-    result, time_taken = _run_pytest(node.path)
+    map_to_runner = {
```
izi-on, line 187: you can do this from the `map_to_handler_run` function: 

`map_to_handler_run = (
    "type",
    {
        "root": (None, _case_root),
        "test": (None, _case_test),
        "conditional": (
            "conditional_type",
            {
                "if": (None, _case_if_node),
            },
        ),
    },
)`

since the map_to_handler_run method has a tuple : (key, dict) where key is the field it will search on the element it has, and dict is the mapping to the handler. In your case, you can add "test_type" and then the dict you have here, then split _case_test into -> _case_test_pytest and _case_test_python. Cleaner since to understand, dev has to read only one dict mapping (map_to_hander_run)

src/renderer/context/testSequencerWS.context.tsx
```diff
@@ -71,7 +73,11 @@ export function TestSequencerWSProvider({
 
   const mapToHandler: { [K in MsgState]: (data: BackendMsg) => void } = {
     TEST_SET_START: (data) => {
-      console.log("starting tests", data);
+      setBackendState("TEST_SET_START");
```
izi-on, line 76: this is just my preference, but if these guys were enums it would be so clean (up to you :) )

src/renderer/routes/test_sequencer_panel/components/TestSequencerInfo.tsx
```diff
@@ -98,14 +105,14 @@ const TestSequencerView = () => {
                   >
                     Save Test Set
                   </LockableButton>
-                  <LockableButton
+                  <Button
                     variant="dotted"
                     className="mt-4 w-full gap-2"
-                    isLocked={_.isEmpty(tree)}
-                    onClick={handleClickRunTest}
+                    disabled={backendState === "TEST_SET_EXPORT"}
```
izi-on, line 111: Quick question: have you considered using the lockable button? During a specific state of the test sequencer, you can chose to have them locked or not without needing to specify the disabled. The thing is, once locked, all the buttons defined as Lockable will also be locked. Just food for thought tho.

izi-on (changes requested):
Amazing stuff, left a few insightful comments. Also, I want @itsjoeoui to review the poetry stuff since he has a lot of experience with it, make sure everything is fine.

captain/utils/test_sequencer/run_test_sequence.py
LatentDream: Oh, it was to debug something. I'll remove it

src/renderer/context/testSequencerWS.context.tsx
LatentDream: Ahah, faaacts

src/renderer/routes/test_sequencer_panel/components/TestSequencerInfo.tsx
LatentDream: It's because I need to Unlock this button even when other are lock. But there is a case where I want to keep it lock (when we are uploading). I'll double check how I can do it

src/renderer/routes/flow_chart/views/ControlBar.tsx
```diff
@@ -44,7 +45,7 @@ const ControlBar = () => {
   const [isDebugSettingsOpen, setIsDebugSettingsOpen] = useState(false);
   const [isEditorSettingsOpen, setIsEditorSettingsOpen] = useState(false);
   const [isDeviceSettingsOpen, setIsDeviceSettingsOpen] = useState(false);
-  const [isDepManagerModalOpen, setIsDepManagerModalOpen] = useState(false);
+  const { isDepManagerModalOpen, setIsDepManagerModalOpen } = useSettingModal();
```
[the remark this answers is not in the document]
LatentDream: Yeah I agree, for now all the "Open Modal" were contain in a single drop down, but as the app grow, we would need to access those from more places. I like the idea of having all modal state variables in one hook, but I'll do some research

src/renderer/context/testSequencerWS.context.tsx
itsjoeoui: In TypeScript let's use a type union instead of an enum :)

Example:
```typescript
// Type union declaration
type MyType = number | string;
```

src/renderer/types/testSequencer.ts
```diff
@@ -8,12 +8,13 @@ export type LockedContextType = {
   isLocked: boolean;
 };
 
-export type TestTypes = "Python" | "Flojoy" | "Matlab";
+export type TestTypes = "Pytest" | "Python" | "Flojoy" | "Matlab";
```
itsjoeoui, line 11: What is the difference between Pytest and just Python here (?)

itsjoeoui (changes requested):

captain/utils/test_sequencer/run_test_sequence.py
LatentDream: Already Type Union. Syed and Joey have strong preference for them

src/renderer/context/testSequencerWS.context.tsx
LatentDream: Already Type Union. Syed and Joey have strong preference for them

src/renderer/types/testSequencer.ts
LatentDream: TEST_SET_START is already for the test Union, and the "Pytest" | "Python" are for the different types as demo in slack

captain/internal/manager.py
LatentDream: Currently, the test sequencer is run as an async function (before having the cancel functionnality) I'm just providing a way to cut the task between tests. A test won't be kill because as Izi pointed out, it's run in a subprocess and the exception raised by stopping the async "queue" won't kill it. 
### Upside:
- It doesn't stop a test, so if a test need a cleanup from pytest, this one will be run.
- Same behavior as the Flowchart
### Down side:
- Won't kill a long running test (Or even a infinit loop)

captain/utils/test_sequencer/run_test_sequence.py
izi-on: > But at the same time I'm duplicating the `_case_test` function :/ Which encapsulate the logic that is the same

Good point, but my "vision" was that the developer would need to create a case only for one type of new test or conditional and just add it to the `map_to_handler_run`. Also, here we breakdown the case for tests in the `_case_test` function, but for the conditional we have the specifics in the `map_to_handler` which is also a small inconsistency. 

So, what I propose as a solution is a functional approach: we wrap whatever is a test in a `handle_test` function in the `map_to_handler` function, the wrapper will basically implement the whole broadcasting process! 

I will handle the code for that, no need for you to make a PR :)

izi-on (approved):
lgtm!

PR #1112, opened 2024-02-26
Sandevh/stu 283 test sequencer drag and drop

izi-on:
- drag and drop for test sequencer (for now, only the rows in the test sequencer)

- future plans: drag and drop conditionals (instead of right click)


itsjoeoui (approved):
lgtm

PR #1114, opened 2024-02-26
Fix: check for version match before marking cached PY interpreter as default

smahmed776:


src/main/python/index.ts
```diff
@@ -28,23 +28,23 @@ export async function checkPythonInstallation(
...
-                ...i,
-                default: true,
-              }
-            : i,
-        );
+      if (foundInterpreterInList) {
+        global.pythonInterpreters = global.pythonInterpreters.map((i) => ({
+          ...i,
+          default: i.path === interpreter ? true : false,
+        }));
```
itsjoeoui, line 43: seems like the `find` and `map` here can just be combined into a single `map`?

itsjoeoui (changes requested):

src/main/python/index.ts
smahmed776: @itsjoeoui I think map and find can not be combined here as Python interpreter retrieved from cache file can possibly not discovered by Python manager. In that case we need to check and push it to pythonInterpreter list.

PR #1118, opened 2024-02-28
STU-270 | Test Sequencer: export complexe data type produced in test

LatentDream:
## Demo
https://github.com/flojoy-ai/studio/assets/56351875/796053e3-2b35-4f5f-abcc-3871d49fdfc4


dstrande (dismissed):
May as well add a docs page in this PR so we don't have to refer back to this video for a how-to. Maybe upload this video to youtube (unlisted) and embed it in the docs? Can do that part later.

LatentDream:
> May as well add a docs page in this PR so we don't have to refer back to this video for a how-to. Maybe upload this video to youtube (unlisted) and embed it in the docs? Can do that part later.

Yes that's a good idea :)

LatentDream:
> One question. How does it know the test_id? If you have a test with multiple different test types?
Oh wait, did I forgot to handle that, let's me test it. Currently, the Test Sequencer retreive the **last** data exported with: 
`test_sequencer.export(...)`. and upload this one. I don't remember what happen when the test expect a bool but we give it a dataframe

dstrande:
> One question. How does it know the test_id? If you have a test with multiple different test types?

Oh does it just look at the function name and look for the test with that name?

LatentDream:
> > One question. How does it know the test_id? If you have a test with multiple different test types?
> 
> Oh does it just look at the function name and look for the test with that name?

The test sequencer tells the test where to store the data based on the testName (in the UI).
[Example of what going on in the background](https://github.com/flojoy-ai/cloud/blob/main/sdks/python/flojoy-cloud/tests/sequencer_export_test.py#L6)

- That way, it works with Pytest, Python and it would works with Matlab in the futur.

izi-on (approved):
lgtm

PR #1119, opened 2024-02-28
App Gallery + Welcome Modal Changes

dstrande:


dstrande:
Added custom app loader to app gallery since it has been removed from the welcome modal.

image

PR #1120, opened 2024-02-28
Iconography

smahmed776:
### ICONOGRAPY
Light theme:
Screenshot 2024-03-05 at 9 21 48 pm

Dark theme:
Screenshot 2024-03-05 at 9 22 04 pm


src/renderer/assets/blocks/block-svg-helper.tsx
```diff
@@ -0,0 +1,16 @@
+import { useTheme } from "@/renderer/providers/themeProvider";
+
+const withThemeSVG = (
+  svgMap: Record<string, (theme?: "dark" | "light") => React.JSX.Element>,
```
itsjoeoui, line 4: Hey Syed, could you please make sure not type the key of a Record as `string`? Since in this case we know what are all the possible keys right

itsjoeoui (changes requested):

src/renderer/assets/blocks/block-svg-helper.tsx
smahmed776: @itsjoeoui I believe using a string type for the key here is appropriate, as the code always checks if the block name is in the map object before loading the value. So, even if you register something unusual in the map object, it won't be selected. And, to be honest, the key is actually the block name, and there are so many blocks that mentioning their names individually would be cumbersome.

src/renderer/components/common/block-label.tsx
```diff
@@ -0,0 +1,34 @@
...
+  variant = "accent2",
+  labelPosition,
+}: BlockLabelProps) => {
+  return (
+    <div className="flex w-full items-center justify-center p-1">
+      <h2
+        style={{ width: textWrap(208, 24, label) }}
+        // FIXME: `text-${variant}` doesn't work for accent6
+        // To test replace `${getVariantClass(variant).text}` with `text-${variant}`
+        // And add an AI block to flowchart
```
39bytes, line 22: Is this still the case? seems to work when testing

39bytes (approved):

PR #1125, opened 2024-03-07
LabView like wire types

smahmed776:
Screenshot 2024-03-08 at 7 16 24 PM
Screenshot 2024-03-08 at 7 17 57 PM


LatentDream (changes requested):
I changed two little thing, just some text box that was not at the right place and two connection that were inverted. 

- **We can't see which node raise an error** when an error occurs, which makes it kinda hard to "debug" and we also can not see the error. Not sure if the regression is from this PR or the clean up one, cc: @39bytes 
- Same for the visual effect of showing which block is currently running. It's not on right now

LatentDream (dismissed):
Previous comment fix here: https://github.com/flojoy-ai/studio/pull/1129

src/renderer/stores/project.ts
```diff
@@ -507,7 +512,27 @@ export const useCreateEdge = () => {
...
+          edge.source === connection.source &&
+          edge.target === connection.target &&
+          edge.sourceHandle === connection.sourceHandle &&
+          edge.targetHandle === connection.targetHandle,
+      );
+      if (edgeExists || !connection.source || !connection.target) {
+        return edges;
+      }
+      const newEdge: Edge<EdgeData> = {
+        id: `${connection.source}->${connection.target}`,
```
39bytes, line 527: this id is not unique per edge, if you have two edges from block A to block B then both of them will have the same id, it's probably better to use a uuid (or do something like ``${connection.source}->${connection.target} ${(connection.sourceHandle)->(connection.targetHandle)}``

docs/src/utils/tailwind.ts
```diff
@@ -0,0 +1,49 @@
+import type { TVariant } from "@/types/tailwind";
+
+export const variantClassMap: Record<
+  TVariant,
+  Record<"stroke" | "fill" | "text" | "shadow" | "border", string>
```
itsjoeoui, line 5: this should be an object instead?

```typescript
type Style = {
    stroke: string
    ...
}
```

itsjoeoui (changes requested):

itsjoeoui (changes requested):
https://github.com/flojoy-ai/studio/assets/25695219/be494bad-5add-4e22-985f-d5ec28efb2a9


Inconsistent resizing behaviour:
- sometimes the frame includes the text label and sometimes it does not (when it is small), this should be consistent

itsjoeoui (changes requested):
https://github.com/flojoy-ai/studio/assets/25695219/29da5b92-743d-4486-a99e-ba8685d19efc

The SVG should also adapt to the scale as you resize it imo, the video is from this reactflow example: https://reactflow.dev/examples/nodes/shapes
I sent the code of this example in Slack a bit ago 👀

src/renderer/stores/project.ts
smahmed776: Nice catch @39bytes . Fixed.

itsjoeoui (approved):
lgtm

PR #1127, opened 2024-03-10
STU-312 fix NI DAQ without chassis error when getting chassis' name

LatentDream:
I'm just complaining, how are we supposed to know that this throws?
```
    @property
    def compact_daq_chassis_device(self):
        """
        :class:`nidaqmx.system.device.Device`: Indicates the name of the
            CompactDAQ chassis that contains this module.
        """

        val = self._interpreter.get_device_attribute_string(self._name, 0x29b7)
        return _DeviceAlternateConstructor(val, self._interpreter)
```

Potential fix for SamChi Pulpan's error in Discord


LatentDream:
* I asked Sam on Discord to run a python script to see with the fix works

LatentDream:
Tested by Sam on Discord, all good !

izi-on (approved):
lgtm

PR #1135, opened 2024-03-16
STU-320: Better notification - Better Modal Management | STU-315: Ability to rename test in UI

LatentDream:
## Changes
- Better error report from pytest to the user
- Refactor some modal where the state was pass left and right
- Loading toast while the tests a discovered and processed
- Rename `Hardware id`  -> `Serial Number`
- STU-315: Ability to rename test in UI


src/renderer/hooks/useTestImport.ts
```diff
@@ -47,15 +47,19 @@ export const useTestImport = () => {
...
       const res = await discoverPytest(path, settings.importAsOneRef);
       if (res.isErr()) {
-        toastQueryError(res.error, "Error while trying to discover tests");
-        return;
+        throw res.error;
       }
       data = res.value;
     }
+    if (data.error) {
+      // Warning: This can be a big trace, might be better not showing it as a toast
```
LatentDream, line 60: Working on creating a Modal service to put all modal at the same place :)

LatentDream:
Found a small detail that matter which is not handled right now, putting back in draft mode

PR #1138, opened 2024-03-21
Test Sequencer | V2

LatentDream:
# Work In Progress

#### UX
- [X] Clearly split Sequencer and Flowchart
- [X] Redesign Sequencer UX

#### Unit Under Test
- [x] UI
- [x] Cloud Integration 

#### Unit Test Environment
- [x] UI
- [x] Backend + Cloud Integration
- [x] Publish automatically

#### Test Sequencer Control
Adding: `Pause`, `Resume`, `Abort`
- [x] UI
- [x] Backend:  `Pause`, `Resume` & `stop`
- [x] Better Visual Indication on what's happening on the test
    - Green, Red, Blue, Yellow ...
- [x] Display global state at the top

#### STU-328 Multi-Sequence
- [x] UI
- [x] Ability to import multiple sequence
- [x] Ability to import **All** test sequences from a folder
- [x] Ability to run those sequence on after the other
- [x] Ability to see there individual status
- [x] Ability to `uncheck` some of them so they don't run
- [x] Ability to save all changes
- [x] STU-333 Migrate sequence import / save to the menu option

#### Cycle
- [x] UI
- [x] Ability to run the **sequencer** in cycle
- [x] Ability to run the **sequencer** indefinitely
- [x] Ability so navigate those cycle

#### Profile Integrity
_Track the integrity of the test session_
- [x] UI
- [x] Backend
    - When a sequence file is unselected
- [x] Publish to cloud with the session

#### Permission
- [x] WIP

#### Test
- [x] STU-287 Playright test suite "Core Functionality"

#### Other
- [X] STU-290 Ability to point to a private Flojoy Cloud


LatentDream:
Depends on: https://github.com/flojoy-ai/cloud/pull/76

src/renderer/routes/test_sequencer_panel/models/models.ts
```diff
@@ -27,24 +27,23 @@ export const testSequenceStopRequest: (
...
-) => {
+) => TestSequenceRun = (tree: TestRootNode) => {
   return {
-    event: "export",
+    event: "pause",
     data: tree,
-    hardware_id: hardware_id,
-    project_id: project_id,
   };
 };
```
39bytes, line 49: these functions feel useless, it seems like less typing to just type out the object literal (`{event: type, data: tree}`) instead of importing these functions and calling them, it should be typesafe either way?

src/renderer/hooks/useTestSequencerState.ts
```diff
@@ -212,26 +266,162 @@ export function useTestSequencerState() {
...
+    // Upload
+    uploadAfterRun,
+    setUploadAfterRun,
+    uploadInfo,
+    handleUpload,
+    setStationId,
+    setSerialNumber,
+    setIntegrity,
+    setIsUploaded,
   };
```
39bytes, line 426: please DO NOT do this
each of the new functions you introduced can just be their own hook (or even just add it as a store action when possible! see `stores/project.ts`)
accessing store state should be as granular as possible

src/renderer/routes/flow_chart/views/env-var/EnvVarModal.tsx
```diff
@@ -103,6 +106,17 @@ const EnvVarModal = () => {
...
+    const validSecret = user.match(
+      () => true,
+      () => {
+        toast.error("Invalid Flojoy Cloud workspace secret");
+        return false;
+      },
+    );
+    if (!validSecret) {
+      return;
+    }
```
39bytes, line 119:
```suggestion
    if (user.isErr()) {
      toast.error("Invalid Flojoy Cloud workspace secret");
      return;
    }
```

src/renderer/stores/sequencer.ts
```diff
@@ -2,36 +2,70 @@ import { v4 as uuidv4 } from "uuid";
...
   setTestSequenceTree: (val: TestRootNode) => void;
-  setTestSequencerProject: (val: TestSequencerProject | null) => void;
+  setUploadAfterRun: (val: boolean) => void;
+  setSerialNumber: (val: string) => void;
+  setIntegrity: (val: boolean) => void;
+  setStationId: (val: string) => void;
+  setIsUploaded: (val: boolean) => void;
+  // Cycles
+  setCycleCount: (val: number) => void;
+  setInfinite: (val: boolean) => void;
```
39bytes, line 53: exposing this many setters as actions does not feel right...
are you sure you can't extract some of the mutations in the components into some composite actions that are easier to understand?
for example, if you find yourself doing `setSequences([])` often, probably just expose something like `clearSequences` instead.

src/renderer/stores/sequencer.ts
```diff
@@ -76,23 +141,360 @@ export const useSequencerStore = create<State & Actions>()(
...
+        if (
+          state.cycleRuns.length > state.cycleConfig.cycleCount &&
+          state.cycleConfig.infinite === false
+        ) {
+          return;
+        }
+        state.cycleRuns.push(state.sequences.map((seq) => ({ ...seq })));
+        state.cycleConfig.ptrCycle = state.cycleRuns.length - 1;
+      }),
+    diplayPreviousCycle: () =>
```
39bytes, line 368: display

src/renderer/stores/sequencer.ts
```diff
@@ -76,23 +141,360 @@ export const useSequencerStore = create<State & Actions>()(
...
+      ? {
+          ...elem,
+          status: StatusType.parse("pending"),
+          completionTime: undefined,
+          isSavedToCloud: false,
+        }
+      : { ...elem };
+  });
+  state.elements = newElems;
+}
```
39bytes, line 500: since these don't use any hooks, you can just make them store actions which can be called by accessing them through `get()`,  then you don't need to pass in the state explicitly. 
alternatively, if you don't want to expose these functions outside of the store, you can get a setter to the state with `useSequencerStore.setState((state) => ...)`

src/renderer/stores/sequencer.ts
```diff
@@ -2,36 +2,70 @@ import { v4 as uuidv4 } from "uuid";
...
+    elements: TestSequenceElement[],
+  ) => void;
+  removeSequence: (name: string) => void;
+  displaySequence: (name: string) => void;
+  runNextRunnableSequence: (sender: SendJsonMessage) => void;
+  runRunnableSequencesFromCurrentOne: (sender: SendJsonMessage) => void;
+  updateSequenceStatus: (val: StatusType) => void;
 };
 
 export const useSequencerStore = create<State & Actions>()(
```
39bytes, line 71: a lot of this store feels like misuse of zustand
if you want to access do this outside of the `set` callback then you can add `get` (`immer((set, get) => ...`) which is just a function that returns the current state.

src/renderer/stores/sequencer.ts
```diff
@@ -76,23 +141,360 @@ export const useSequencerStore = create<State & Actions>()(
...
+              if (
+                el.type === "test" &&
+                (el.status === "running" || el.status === "paused")
+              ) {
+                el.status = "aborted";
+              }
+            });
+          }
+        }
+      }),
```
39bytes, line 218: can you avoid putting the entire logic of the function inside of the `set` callback?
similarly to try/catch, only put stuff that actually mutates the state inside of `set` (here you have extra validation logic that doesn't belong in lines 200-205)
you can use `get()` to access the state.
it's better because it clearly marks where mutations are happening so it's easier to read/debug.

src/renderer/stores/sequencer.ts
```diff
@@ -76,23 +141,360 @@ export const useSequencerStore = create<State & Actions>()(
...
+          );
+          if (oldIdx !== -1) {
+            // Could be -1 if the sequence was removed
+            const oldSequence = containerizeCurrentSequence(oldIdx, state);
+            state.sequences[oldIdx] = oldSequence;
+          }
+        }
+        // Load the new sequence
+        loadSequence(idx, state);
+      }),
```
39bytes, line 339: same comment with the rest of all of these actions (and anywhere it happens in the rest of the store too), limit the scope of `set`.

src/renderer/routes/test_sequencer_panel/components/modals/RenameTestModal.tsx
```diff
@@ -0,0 +1,62 @@
+import { Button } from "@/renderer/components/ui/button";
+import { Dialog, DialogContent } from "@/renderer/components/ui/dialog";
+import { Input } from "@/renderer/components/ui/input";
+import { useModalState } from "@/renderer/hooks/useModalState";
+import { useTestSequencerState } from "@/renderer/hooks/useTestSequencerState";
+import { useState, useEffect } from "react";
+
+export const RenameTestModal = () => {
+  const { renameTestId, isRenameTestModalOpen, setIsRenameTestModalOpen } =
+    useModalState();
```
39bytes, line 10: should be passed as props

src/renderer/routes/test_sequencer_panel/components/modals/ModalsProvider.tsx
```diff
@@ -0,0 +1,15 @@
...
+export function ModalsProvider() {
+  return (
+    <div>
+      <TestSequencerProjectModal />
+      <ImportTestModal />
+      <ErrorModal />
+      <RenameTestModal />
+    </div>
+  );
+}
```
39bytes, line 15: bad naming, this isn't a provider because it doesn't supply any context (that's what `Provider` components always do in React)
also since this is only used in a single place, it's kind of a useless component anyway, just put the 4 modals in the sequencer panel page.

src/renderer/routes/test_sequencer_panel/components/modals/ImportTestModal.tsx
```diff
@@ -16,12 +17,8 @@ export type ImportTestSettings = {
...
 export type ImportType = "pytest" | "python";
 
-type Props = {
-  isModalOpen: boolean;
-  handleModalOpen: Dispatch<SetStateAction<boolean>>;
-};
-
-export const ImportTestModal = ({ isModalOpen, handleModalOpen }: Props) => {
+export const ImportTestModal = () => {
+  const { isImportTestModalOpen, setIsImportTestModalOpen } = useModalStore();
```
39bytes, line 21: these were passed as props before and they should stay that way

src/renderer/routes/test_sequencer_panel/components/modals/ErrorModal.tsx
```diff
@@ -0,0 +1,31 @@
+import { Dialog, DialogContent } from "@/renderer/components/ui/dialog";
+import { useModalState } from "@/renderer/hooks/useModalState";
+
+export const ErrorModal = () => {
+  const { isErrorModalOpen, setIsErrorModalOpen, errorModalMessage } =
+    useModalState();
```
39bytes, line 6: pass as props instead

src/renderer/routes/test_sequencer_panel/components/dnd/DraggableRowSequence.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { TestSequenceContainer } from "@/renderer/types/test-sequencer";
+
+export const DraggableRowSequence = ({
+  isSelected,
+  row,
+  ...props
+}: {
+  isSelected: boolean;
+  row: Row<TestSequenceContainer>;
+}): React.ReactNode => {
```
39bytes, line 20: no need to mark return type
props type shouldn't be inlined for style reasons (do `type Props = { ... }`)
useless spread operator (there's only 2 props and you already destructured them both, `...props` is completely empty)

src/renderer/routes/test_sequencer_panel/components/DesignBar.tsx
```diff
@@ -0,0 +1,266 @@
...
+    <div className="inline-flex h-full w-full items-center justify-center rounded-md bg-red-500 text-primary-foreground">
+      <code data-testid="global-status-badge">ABORTED</code>
+    </div>
+  ),
+  fail: (
+    <div className="inline-flex h-full w-full items-center justify-center rounded-md bg-red-500 text-primary-foreground">
+      <code data-testid="global-status-badge">FAIL</code>
+    </div>
+  ),
+};
```
39bytes, line 259: this could be a lot shorter by turning it into its own component and cutting out a lot of the repeated css (lookup `class-variance-authority`, it's already installed in studio)

src/renderer/routes/test_sequencer_panel/components/DesignBar.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+const getNumberOfSequenceRun = (data: TestSequenceContainer[]): number => {
+  let count = 0;
+  data.forEach((elem) => {
+    if (elem.runable && elem.status != "pending" && elem.status != "running")
+      count++;
+  });
+  return count;
+};
```
39bytes, line 191: too much repeating logic
use this `countWhere` function instead and define all of these functions in terms of `countWhere` by just changing what you pass to predicate
```ts
export function countWhere<T>(arr: T[], predicate: (t: T) => boolean): number {
  return arr.reduce((acc, t) => (predicate(t) ? acc + 1 : acc), 0);
}
```

src/renderer/routes/test_sequencer_panel/components/DesignBar.tsx
```diff
@@ -0,0 +1,266 @@
...
+          .reduce((prev, curr) =>
+            priority[prev] > priority[curr] ? prev : curr,
+          )
+      : "pending";
+  if (sequences.length === 0 && data.length === 0) return highestCycle;
+  // Highest in the view
+  const iter = sequences.length > 0 ? sequences : data;
+  const status = iter
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .map((el: any) => el.status)
```
39bytes, line 221: whenever you find yourself going to `any` you should ask yourself if you're doing something wrong.
in this case, `TestSequenceElement` could possibly be a `Conditional` which does not have a status field.
you should just filter only the `Test`s in `data` so then you can type this properly (`TestSequenceContainer | Test`)

src/renderer/routes/test_sequencer_panel/components/DesignBar.tsx
```diff
@@ -0,0 +1,266 @@
...
+  useEffect(() => {
+    setSummary({
+      numberOfTestRun: getNumberOfTestRun(elems),
+      numberOfTest: getNumberOfTest(elems),
+      numberOfSequence: getNumberOfSequence(sequences),
+      numberOfSequenceRun: getNumberOfSequenceRun(sequences),
+      successRate: getSuccessRate(elems),
+      status: getGlobalStatus(cycleRuns, sequences, elems),
+    });
+  }, [elems, sequences, cycleRuns]);
```
39bytes, line 57: since all of this is derived from elems, sequences, and cycleRuns, there's 0 reason for this to be held as state at all
just use `useMemo` if you only want to recalculate it when the dependencies change.

src/renderer/routes/test_sequencer_panel/components/data-table/utils.tsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export const getCompletionTime = (data: TestSequenceElement[]) => {
+  const onlyTests = getOnlyTests(data);
+  const parallel = filter(onlyTests, (elem) => elem.runInParallel).map(
+    (elem) => elem.completionTime,
+  );
+  const nonParallel = filter(onlyTests, (elem) => !elem.runInParallel).map(
+    (elem) => elem.completionTime,
+  );
```
39bytes, line 81: use `_.partition`

src/renderer/routes/test_sequencer_panel/components/data-table/utils.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const getOnlyTests = (data: TestSequenceElement[]): Test[] => {
+  return filter(
+    data,
+    (elem) =>
+      elem.type === "test" &&
+      elem.status != "pending" &&
+      elem.status != "running",
+  ) as Test[];
+};
```
39bytes, line 96: define a type guard instead of using an `as` cast (essentially does the same thing but it's more idiomatic)
```ts
const isTest = (data: TestSequenceElement): data is Test => {
  return elem.type === "test"
}
```
then you can use this in the filter predicate and typescript should infer the type for you (i think)

src/renderer/routes/test_sequencer_panel/components/ControlButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  };
+
+  const handleClickResumeTest = () => {
+    toast.info("Resuming sequencer.");
+    if (backendGlobalState === "test_set_start") {
+      console.log("Resume test");
+      tSSendJsonMessage(testSequenceResumeRequest(tree));
+    }
+  };
```
39bytes, line 45: these are probably things that should be exposed in a hook defined in the sequencer store file

src/renderer/routes/test_sequencer_panel/components/CloudPanel.tsx
```diff
@@ -13,18 +13,73 @@ import {
...
+  const { isAdmin } = useWithPermission();
+  const [lotNumber, setLotNumber] = useState("");
+  const [description, setDescription] = useState("N/A");
   const [projectId, setProjectId] = useState("");
-  const { tree, setIsLocked } = useTestSequencerState();
-  const { tSSendJsonMessage } = useTestSequencerWS();
+  const [partNumber, setPartNumber] = useState("");
+  const [partVarId, setPartVarId] = useState("");
+  const [serialNumbers, setSerialNumbers] = useState<string[]>([]);
+  const [units, setUnits] = useState<Map<string, Unit>>(new Map());
```
39bytes, line 47: it's not worth it to use `Map` with immutable state, just use a `Record` so you get spread syntax

src/renderer/routes/flow_chart/views/env-var/EnvVarModal.tsx
```diff
@@ -121,6 +135,43 @@ const EnvVarModal = () => {
...
+    const validUrl = serverHealth.match(
+      () => true,
+      () => {
+        toast.error("Invalid Flojoy Cloud URL");
+        return false;
+      },
+    );
+    if (!validUrl) {
+      return;
+    }
```
39bytes, line 156:
```suggestion
    if (serverHealth.isErr()) {
      toast.error("Invalid Flojoy Cloud URL");
      return;
    }
```

src/renderer/hooks/useModalState.ts
```diff
@@ -0,0 +1,37 @@
+import { useModalStore } from "@/renderer/stores/modal";
```
39bytes, line 1: modal store isn't really needed so you can get rid of this mega hook too

src/renderer/context/testSequencerWS.context.tsx
```diff
@@ -81,34 +98,37 @@ export function TestSequencerWSProvider({
...
-
     const msg = BackendMsg.safeParse(lastJsonMessage);
     if (!msg.success) {
       console.error(msg.error);
       return;
+    } else {
+      console.log("Received message from backend: ", msg.data);
     }
 
     switch (msg.data.state) {
```
39bytes, line 109: questionable, each of these handlers could just be a store action (alternatively just turn the entire switch case into a store action that takes in a socket message)

39bytes (changes requested):

src/renderer/routes/test_sequencer_panel/components/DesignBar.tsx
LatentDream: It's because I'm relying on the "duck typing" of sequences, and TestElems, both of them having the status field, might not be the best way to do it since if we change the field in one of those objects, it will break without notice. I'll change it

src/renderer/routes/test_sequencer_panel/components/CloudPanel.tsx
LatentDream: `Map` is the right element here since I don't know all the possible key, the facts that it's a state is a problem tho.
- Converted to a simple map variable

captain/internal/manager.py
```diff
@@ -22,31 +24,80 @@ def __init__(self):
...
+    def new_runner(self, runner: asyncio.Runner, *args, **kwargs):
+        if self.runner is not None:
+            self.kill_runner()
+        self.runner = runner
+        self.pause = False
+        self.poison_pill = None
+
+    def cleanup(self, *args, **kwargs):
+        self.runner = None
+        self.pause = False
```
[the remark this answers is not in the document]
LatentDream: leftover from when upload was handle at the same time. Removing it

captain/internal/manager.py
```diff
@@ -22,31 +24,80 @@ def __init__(self):
...
 
+    def pause_runner(self, *args, **kwargs):
+        if self.runner is not None:
+            logger.info("Pausing TS Runner")
+            self.pause = True
+
+    def resume_runner(self, *args, **kwargs):
+        if self.pause:
+            logger.info("Resuming TS Runner")
+            self.pause = False
```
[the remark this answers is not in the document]
LatentDream: see above, remving it

captain/utils/test_sequencer/handle_data.py
```diff
@@ -3,14 +3,14 @@
...
 from captain.utils.test_sequencer.run_test_sequence import (
-    export_test_sequence,
     run_test_sequence,
 )
 from typing import Callable
 from threading import Lock
 from captain.utils.config import ts_manager
 
 lock = Lock()
+global ts_pause
```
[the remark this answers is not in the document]
LatentDream: nice catch, from a test I did

src/main/utils.ts
```diff
@@ -152,6 +152,84 @@ export const openFilePicker = (
   });
 };
 
+export const openFilesPicker = (
+  _,
+  allowedExtensions: string[] = ["json"],
+  title: string = "Select Files",
+): Promise<{ filePath: string; fileContent: string }[] | undefined> => {
+  // Return mutiple files or all file with the allowed extensions if a folder is selected
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
LatentDream: Yeah, The functions in this file alway use the `return new Promise((resolve, reject) =>` with  `showOpenDialogSync` just a different way to write it. I can change them

src/renderer/routes/test_sequencer_panel/components/CloudPanel.tsx
39bytes: `Map` and `Record` do basically the same thing though, it would just be `Record<string, Unit>` which can easily be updated immutably

src/renderer/routes/common/Layout.tsx
```diff
@@ -1,67 +1,21 @@
...
-    })),
-  );
-
-  const { activeTab } = useAppStore(
-    useShallow((state) => ({
-      activeTab: state.activeTab,
-    })),
-  );
+export const LAYOUT_TOP_HEIGHT_FLOWCHART =
+  LAYOUT_TOP_HEIGHT + SERVER_STATUS_HEIGHT + 32;
```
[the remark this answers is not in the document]
LatentDream: A + 32 and a - 32 elsewhere, removed both

src/renderer/routes/common/Layout.tsx
```diff
@@ -1,67 +1,21 @@
...
-export const Layout = () => {
-  const serverStatus = useSocketStore((state) => state.serverStatus);
-
-  const { setProjectName, projectName, hasUnsavedChanges } = useProjectStore(
-    useShallow((state) => ({
-      setProjectName: state.setProjectName,
-      projectName: state.name,
-      hasUnsavedChanges: state.hasUnsavedChanges,
-    })),
-  );
```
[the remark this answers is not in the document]
LatentDream: Shoooot, I forget to add it back

src/renderer/stores/modal.ts
```diff
@@ -0,0 +1,64 @@
+import { create } from "zustand";
```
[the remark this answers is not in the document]
LatentDream: Canceling this comment - as discussed on Slack. The state as been deleted also

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -1,11 +1,10 @@
...
+// Take the sequence as an Input.
+//   1. Always valide the state of the sequence
 //   2. Do the operation
 //   3. Update the test sequencer
 //   4. Return a Result<T, Error>
-// Note: All path in a project should be using '/' for cross-platform compatibility
-
+// Note: All path in a sequence should be using '/' for cross-platform compatibility
 // Note: Toast.promise doesn't work with neverthrow -----
 export type Result<T, E = Error> =
```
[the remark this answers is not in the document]
LatentDream: Nop, there is a problem with `Toast.promise` & `neverthrow`. I don't remember exactly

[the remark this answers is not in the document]
LatentDream: Nvermind, swithcing for your `toastResultPromise` in `report-error.ts` fix the issue

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -20,135 +19,174 @@ export function Err<T, E>(error: E): Result<T, E> {
...
-  setUnsaved: (unsaved: boolean) => void;
+  sequences: TestSequenceContainer[];
 };
 
-export async function createProject(
-  project: TestSequencerProject,
+export async function createSequence(
+  sequence: TestSequencerProject,
   stateManager: StateManager,
   throwInsteadOfResult: boolean = false,
```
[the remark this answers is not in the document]
LatentDream: Removed the `throwInsteadOfResult` arg of each exposed function

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -162,78 +200,87 @@ function buildResultFromCatch(e: unknown): Result<null, Error> {
   if (e instanceof Error) {
     return Err(e);
   } else {
-    console.error("[SaveProject] Unknown error", e);
-    return Err(new Error("Unknown error while creating the project"));
+    console.error("[Save Sequence] Unknown error", e);
+    return Err(new Error("Unknown error while creating the sequence."));
   }
 }
```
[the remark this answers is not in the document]
LatentDream: Huuummm, each exposed fonction would be:
```
export function createSequence(...): Result<...> {
    return fromThrowable(() -> {
       ... actual function
     })
```
Instead of a try catch, kinda ugly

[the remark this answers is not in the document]
LatentDream: (I'm not a fan of arrow function everywhere ahah)

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -20,135 +19,174 @@ export function Err<T, E>(error: E): Result<T, E> {
...
-    await saveToDisk(project);
-    await syncProject(project, stateManager);
+    await saveToDisk(sequence);
+    await syncSequence(sequence, stateManager);
     return Ok(null);
   } catch (e: unknown) {
     if (throwInsteadOfResult) throw e;
     return buildResultFromCatch(e);
   }
 }
```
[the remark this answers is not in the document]
LatentDream: I don't agree in this case, this file uses the exception, it's actively throws exception as a way to cut the exception. I know we're not active fans of the the use exception now, but in this case I'd say changing it would make the file less readable.

The exposed API doesn't throw, but the inner logic does as a way to make the hole logic easy to understand

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -278,6 +325,7 @@ async function createTestSequencerElementsFromProjectElements(
           ...elem,
         };
   });
+  sequence.elems = elements;
   if (verifStateOrThrow) {
     await throwIfNotInAllBaseFolder(elements, baseFolder);
```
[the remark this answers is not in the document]
LatentDream: Same as previous comment, hole file actively use Exception. But the exposed function are wrap in the neverthrow' Result type

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -20,135 +19,174 @@ export function Err<T, E>(error: E): Result<T, E> {
...
+export async function verifyElementCompatibleWithSequence(
+  sequence: TestSequencerProject,
   elements: TestSequenceElement[],
   throwInsteadOfResult: boolean = false,
 ): Promise<Result<null, Error>> {
-  // Verify that the elements are within the current project directory.
+  // Verify that the elements are within the current sequence directory.
   try {
-    await throwIfNotInAllBaseFolder(elements, project.projectPath);
+    await throwIfNotInAllBaseFolder(elements, sequence.projectPath);
```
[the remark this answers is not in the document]
LatentDream: Same as previous comment

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -162,78 +200,87 @@ function buildResultFromCatch(e: unknown): Result<null, Error> {
...
+  const splitName = name.split("/");
+  const splitBase = new Set(baseFolder.split("/"));
+  const filtered = splitName.filter((el) => !splitBase.has(el));
+  return filtered.join("/");
+}
+
+async function createExportableSequenceElementsFromTestSequencerElements(
   elems: TestSequenceElement[],
   baseFolder: string,
   verifStateOrThrow: boolean,
```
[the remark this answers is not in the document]
LatentDream: Just a way to easily understand that this function throw. The name is long enough ahah

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -1,11 +1,10 @@
...
-//   1. Always valide the state of the proejct
+// Take the sequence as an Input.
+//   1. Always valide the state of the sequence
 //   2. Do the operation
 //   3. Update the test sequencer
 //   4. Return a Result<T, Error>
-// Note: All path in a project should be using '/' for cross-platform compatibility
-
+// Note: All path in a sequence should be using '/' for cross-platform compatibility
 // Note: Toast.promise doesn't work with neverthrow -----
```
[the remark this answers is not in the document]
LatentDream: Done

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -1,11 +1,10 @@
...
-// Take the project as an Input.
-//   1. Always valide the state of the proejct
+// Take the sequence as an Input.
+//   1. Always valide the state of the sequence
 //   2. Do the operation
 //   3. Update the test sequencer
 //   4. Return a Result<T, Error>
-// Note: All path in a project should be using '/' for cross-platform compatibility
-
+// Note: All path in a sequence should be using '/' for cross-platform compatibility
```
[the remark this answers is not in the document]
LatentDream: Converted exposed function to only return a nverthrow `result` but as previous comment, let's not touch the inner logic

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
39bytes: actually if we're not throwing anything anywhere we don't need to use `fromThrowable` at all, either just return errors early or chain together `andThen`

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -0,0 +1,356 @@
...
+  newPath: string,
+): TestSequencerProject {
+  return {
+    ...sequence,
+    projectPath: newPath,
+  };
+}
+
+async function saveToDisk(sequence: TestSequencerProject): Promise<void> {
+  if ("api" in window) {
```
39bytes, line 204: you don't need to check if api is in window everywhere, we're always running within the electron context

src/renderer/routes/test_sequencer_panel/components/data-table/utils.tsx
LatentDream:
```
const isTest = (data: TestSequenceElement): data is Test => {
  return data.type === "test"
}

export const getOnlyCompletedTests = (data: TestSequenceElement[]): Test[] => {
  return filter(
    data,
    (elem) =>
      isTest(elem) &&
      elem.status != "pending" &&
      elem.status != "running",
  );
};
```
Nop, this .type === "test" is quit anoying

src/renderer/routes/test_sequencer_panel/components/ControlButton.tsx
LatentDream: I'm migrating them to the sequencerState, so it's at the same place of the `runSequencer`

src/renderer/routes/test_sequencer_panel/models/models.ts
LatentDream: Typing problem otherwise since the socket can take more than a single type of message

LatentDream: We could expose a function just to send this type of msg tho. I'll look at it tmr while updating the `sequencer.context.ts`

src/renderer/hooks/useTestSequencerState.ts
LatentDream: I splitted this in two separated hooks, + removed all that were just "pass through". Those are now access with `sequencerStore(useShallow(state) => ...)`

LatentDream: What left in the new SequencerState is:
- Run, Abort, Pause, Resume
- Upload
- AddNewSequence
- RemoveSequence (this one just so it's consistent with AddNewSequence being access through there)

src/renderer/stores/sequencer.ts
LatentDream: I know, but it's kind of hard to do without them. Some are necessary for things like the draggable row (like `setSequences(...)`) for example, and others are there to parameterize execution.

src/renderer/stores/sequencer.ts
LatentDream: I don't want to expose them, but I the `useSequencerStore.setState((state) => ...)`, right now we don't know which one need a setter and which one is ok with a `get()`
I'll update them

src/renderer/context/testSequencerWS.context.tsx
LatentDream: The switch statement will have to be redone at the same time as the state management, changing this makes things more complicated and confusing, because as it uses other hooks, half the logic is somewhere, and the other half is somewhere else.
I should have changed everything at the beginning instead of building on what already existed, but I didn't have the time due to the big scope / short deadlines

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
LatentDream: WIP on refactoring the file to remove exception for neverthrow [here](https://github.com/LatentDream/flojoy-studio/pull/5/files)

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
LatentDream: WIP on refactoring the file to remove exception for neverthrow [here](https://github.com/LatentDream/flojoy-studio/pull/5/files)

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
LatentDream: Done :)

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
LatentDream: Merged

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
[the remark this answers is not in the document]
LatentDream: Merged :)

src/renderer/hooks/useTestImport.ts
```diff
@@ -40,22 +43,25 @@ export const useTestImport = () => {
...
       const res = await discoverPytest(path, settings.importAsOneRef);
       if (res.isErr()) {
-        toastQueryError(res.error, "Error while trying to discover tests");
-        return;
+        throw res.error;
       }
       data = res.value;
     }
+    if (data.error) {
+      throw new Error(data.error);
```
LatentDream, line 63: This defeats the purpose of using neverthrow, I'm changing it at the moment

39bytes (commented):
few tiny things, mostly good though
mainly just trying to get this merged quickly so that we can push out a new release
i trust that we'll go back and clean up the tech debt when we have some more time

src/renderer/hooks/useTestSequencerProject.ts
```diff
@@ -1,110 +1,120 @@
...
+      elems,
+      addNewSequence,
+      removeSequence,
       project,
+      sequences,
     };
   }
-  return { setElems, setProject, setUnsaved, elem: elems, project };
+  return { elems, project, addNewSequence, removeSequence, sequences };
 }
```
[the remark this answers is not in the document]
LatentDream: Yeah, you are right, I'll just double check that the change doesn't give a unwanted permission

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -0,0 +1,330 @@
...
+  }
+  const elems = stateManager.elems;
+  sequence = validatePath(sequence);
+  const exportableElems =
+    await createExportableSequenceElementsFromTestSequencerElements(
+      elems,
+      sequence.projectPath,
+    );
+  if (exportableElems.isErr()) {
+    return err(exportableElems.error);
```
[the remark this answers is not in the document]
LatentDream: Some yeah, but some doesn't have the same value type

PR #1148, opened 2024-04-10
Min size window

itsjoeoui:


LatentDream (dismissed):
lgtm, I'll re-work the control bar a bit later so we can allow the app to have a min width a bit smaller later today

LatentDream (approved):

PR #1150, opened 2024-04-11
[STU-343] Edit sequence name & description

LatentDream:
- [X] Rename sequence name
- [X] Edit description
- [X] Fix: `You have unsaved changes` when removing a sequence after a `Ctrl+S` 
  - The "unsaved" field might be better in the "Sequence" object instead of the "SequenceContainer" for a cleaner implementation. 
  - Taking note of this one for the refactoring of the state management.


src/renderer/routes/test_sequencer_panel/components/modals/RenameModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  target,
+}: {
+  title: string;
+  isModalOpen: boolean;
+  setModalOpen: (value: boolean) => void;
+  initialName: string;
+  handleSubmit: (newName: string) => void;
+  target: string;
+}) => {
+  const [newName, setNewName] = useState<string>(initialName);
```
itsjoeoui, line 21: I don't think this state here is necessary (?)

itsjoeoui (changes requested):

src/renderer/routes/test_sequencer_panel/components/modals/RenameModal.tsx
39bytes: it is, it's a form, you don't want to commit the changes until hit the confirm button

src/renderer/routes/test_sequencer_panel/components/data-table/SequenceTable.tsx
```diff
@@ -255,8 +258,82 @@ export function SequenceTable() {
...
+        if (idx === renameForIdx.current) {
+          return {
+            ...seq,
+            project: { ...seq.project, description: newDescription },
+            testSequenceUnsaved: true,
+          };
+        }
+        return seq;
+      }),
+    );
```
39bytes, line 315: consider using immer whenever you need to modify nested properties, it's a lot more readable
```typescript
setSequences(produce(sequences, (draft) => {
  const seq = draft[renameForIdx.current];
  seq.project.description = newDescription;
  seq.testSequenceUnsaved = true;
}))
```

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -82,6 +83,15 @@ export async function saveSequence(
   if (isSync.isErr()) {
     return isSync;
   }
+  // Set the sequence as saved
+  stateManager.setSequences(
+    [...stateManager.sequences].map((seq) => {
```
39bytes, line 88: you don't need to spread here i think

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -110,6 +120,12 @@ export async function saveSequences(
     if (res.isErr()) {
       return err(res.error);
     }
+    stateManager.setSequences(
+      [...stateManager.sequences].map((seq) => ({
```
39bytes, line 124: no need to spread (?)

src/renderer/routes/test_sequencer_panel/components/modals/RenameModal.tsx
itsjoeoui: ah right I see I see

itsjoeoui (approved):

src/renderer/routes/test_sequencer_panel/components/data-table/SequenceTable.tsx
LatentDream: Oh that's wayyyy better

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
LatentDream: True, but I realized that the change is not reflected right away in the UI :/

itsjoeoui (approved):

39bytes (approved):

PR #1155, opened 2024-04-13
[STU-344 & STU-345 & STU-347] cloud info

LatentDream:
- [x] A bit of cleanup and standardisation in the file where we call Flojoy Cloud
- [x] Removed the Station-ID from UI since the info is implicite with the Station
- [x] Adding the Product ID in for the description of the part under test
- [x] Colorous background for the tests result.

### DIFF
--- a/src/renderer/global.css
+++ b/src/renderer/global.css
@@ -63,7 +63,6 @@
     --warning-bg: hsl(49, 100%, 97%);
     --warning-border: hsl(49, 91%, 91%);
     --warning-text: hsl(31, 92%, 45%);
-      
   }
 
   .dark {
@@ -108,7 +107,7 @@
     --color-accent4: 240 95 168;
     --color-accent5: #8c9599;
     --color-accent6: #88fa4e;
-    
+
     --success-bg: hsl(150, 100%, 6%);
     --success-border: hsl(147, 100%, 12%);
     --success-text: hsl(150, 86%, 65%);
--- a/src/renderer/global.css
+++ b/src/renderer/global.css
@@ -47,6 +47,23 @@
     --color-accent4: 240 95 168;
     --color-accent5: #8c9599;
     --color-accent6: #1db100;
+
+    --success-bg: hsl(143, 85%, 96%);
+    --success-border: hsl(145, 92%, 91%);
+    --success-text: hsl(140, 100%, 27%);
+
+    --info-bg: hsl(208, 100%, 97%);
+    --info-border: hsl(221, 91%, 91%);
+    --info-text: hsl(210, 92%, 45%);
+
+    --error-bg: hsl(359, 100%, 97%);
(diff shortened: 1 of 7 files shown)


LatentDream:
# Preview of Background for the test result
dark
light

src/renderer/routes/test_sequencer_panel/components/dnd/DraggableRowStep.tsx
```diff
@@ -71,10 +75,34 @@ export const DraggableRowStep = ({
...
+  let variant = "default" as
+    | "aborted"
+    | "pass"
+    | "fail"
+    | "running"
+    | "paused"
+    | "default";
+  if (row.original.type === "test") {
+    if (row.original.status !== "pending") {
+      variant = row.original.status as
```
LatentDream, line 87: @39bytes, I'm not sure about `as`, but it's the best way I've found to convert a status to a variant. + there's a check to convert to default if the value doesn't exist (and Typescript will complain if a new status is added).
Curious if you know of any TS magic for this?

src/renderer/global.css
```diff
@@ -47,6 +47,22 @@
     --color-accent4: 240 95 168;
     --color-accent5: #8c9599;
     --color-accent6: #1db100;
+
+    --success-bg: hsl(143, 85%, 96%);
```
itsjoeoui, line 51: 2 suggestions, pick one of them :)

1. Add those to `tailwind.config.js`, you can follow the same approach as other colors such as `primary`, `secondary`, `destructive`, etc. This is better because the LSP gives you auto complete.
2. Just add tailwind colors to variant in `table.tsx` instead of using `hsl`

itsjoeoui (changes requested):

src/renderer/routes/test_sequencer_panel/components/dnd/DraggableRowStep.tsx
LatentDream: > I'm confused as to why the .includes check is there when the type of row.original.status inside the if block is "aborted" | "pass" | "fail" | "running" | "paused"? 

Rigth now, it's never hit, it's only for the linter (+ let's say you add something else, the runtime don't panick, but typescript will tell us, so it's not actually needed). That's why I'm asking.

I'm testing your solutions, thanks

LatentDream:
```
  type Variant = Exclude<StatusType, "pending"> | "default";
  let variant: Variant = "default";
  if (row.original.type === "test") {
    if (row.original.status !== "pending") {
      variant = row.original.status;
    }
  }
```
The Exclude is perfect, + Typescript signals us that something is wrong when adding a new value in StatusType. Thanks

itsjoeoui (approved):

PR #1156, opened 2024-04-16
[STU-337] Cloud integration: Loading test profile 

LatentDream:
# Description
If a repo is specify with a Test Profile, provide the user with the ability to automatically load the all the sequences attached to this one. 
- Added a fix for the Upload (an Alias was missing for the conversion `camelCase` -> `snake_case`

# TODO list
- [x] If URL is specify load the Test Profile
   - Viewer, will load automatically
   - Admin, will ask confirmation
- [x] Production Line -> Test Profile (to have the same name as the mock)
- [x] Viewer permission
- [x] Having an example with the default production line in Cloud

# Some warning
- Currently assume Git is installed
- Private repo works only if the user has set git credential

# To test
1. Create a repo with a sequence(s) (.tjoy) saved at the root.
2. Put the repo URL in the Test Profile in Cloud
3. Select the Test Profile in the sequencer
- Everything will be automatically loaded


src/renderer/lib/api.ts
```diff
@@ -179,6 +179,10 @@ export type Part = z.infer<typeof Part>;
 const Project = z.object({
   label: z.string(),
   value: z.string(),
+  repoUrl: z
```
itsjoeoui, line 182: maybe `z.string().default("")` (?)

itsjoeoui (changes requested):

src/renderer/hooks/useTestSequencerState.ts
```diff
@@ -420,6 +452,15 @@ export function useSequencerState() {
     }
   }
 
+  function clearSequencer() {
+    if (isLocked) {
+      toast.error("Cannot clear sequencer while running.");
+      return;
+    } else {
+      clearState();
+    }
```
39bytes, line 461: `else` not required

src/renderer/lib/api.ts
```diff
@@ -228,7 +232,7 @@ export const postSession = (
             cycleNumber: cycleNumber,
             name: elem.testName,
             pass_: elem.status === "pass",
-            completionTime: elem.completionTime!,
+            completionTime: elem.completionTime ? elem.completionTime : 0,
```
39bytes, line 235:
```suggestion
            completionTime: elem.completionTime ?? 0,
```

src/renderer/lib/api.ts
LatentDream: default is to replace `undefined` only, not null

PR #1159, opened 2024-04-17
Stu-346 placeholder test

LatentDream:
# Description
The goal is to have the ability to define test steps without any code first. For that, a `placeholder` test type is added. This one can have Min / Max / Unit + a name. When the coder is ready to code, he can link it to a real Pytest or Python script.

# TODO Features
- [x] Placeholder Type
- [x] Placeholder Runner
- [x] Change to what a test points to
- [x] Give the ability to already assign tests to be re-assigned

# TODO Before merge
- [x] CSS overflow when test names are long
- [x] Make sure it works with loaded sequences
- [x] Test Permission


src/renderer/routes/test_sequencer_panel/components/modals/CreatePlaceholderTestModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+        undefined,
+        undefined,
+        min,
+        max,
+        unit,
+      ),
+    ]);
+    if (res.isErr()) {
+      return;
+    }
```
39bytes, line 44: no error toast?

src/renderer/routes/test_sequencer_panel/components/modals/ChangeLinkedTest.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    );
+                  }
+                })}
+              </SelectGroup>
+            </SelectContent>
+          </Select>
+          <div className="w-52 justify-end">
+            <Button
+              variant={"outline"}
+              onClick={() => handleDiscoverPytestElements()}
```
39bytes, line 122:
```suggestion
              onClick={handleDiscoverPytestElements}
```

src/renderer/routes/test_sequencer_panel/components/modals/CreatePlaceholderTestModal.tsx
LatentDream: The Validator in AddNewElements already display the Toast :)

src/renderer/routes/test_sequencer_panel/components/modals/CreatePlaceholderTestModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [unit, setUnit] = useState("");
+
+  const { addNewElems } = useDisplayedSequenceState();
+
+  const handleCreate = async () => {
+    if (name === "") {
+      toast.error("Please enter a test name");
+    }
+    const res = await addNewElems([
+      createNewTest(
```
itsjoeoui, line 30: maybe consider take the params as an object? This makes it a bit hard to understand what each parameter is for :(

src/renderer/routes/test_sequencer_panel/components/modals/CreatePlaceholderTestModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <h2 className="text-lg font-bold text-accent1">
+          Create a placeholder test step
+        </h2>
+        <p className="text-xs text-muted-foreground">
+          {" "}
+          This will create a new test step with the given name and expected
+          value without being link to any code. The code can be added later.
+        </p>
+
+        <div className="grow pb-1 text-xs">
```
itsjoeoui, line 60: I would actually recommend using a form to handle those input states, which will also let you do validation easily.

itsjoeoui (dismissed):

src/renderer/hooks/useTestSequencerState.ts
```diff
@@ -98,34 +99,38 @@ const validateElements = (
...
+    testName: test.name,
     runInParallel: false,
-    testType: type,
+    testType: test.type,
     status: "pending",
     completionTime: undefined,
     error: null,
     isSavedToCloud: false,
-    exportToCloud: exportToCloud === undefined ? true : exportToCloud,
+    exportToCloud: test.exportToCloud || true,
```
39bytes, line 129: `|| true` is redundant, could just be `true` (though is this intentional? why not just `test.exportToCloud`)

src/renderer/routes/test_sequencer_panel/components/modals/ChangeLinkedTest.tsx
```diff
@@ -0,0 +1,168 @@
...
+        setAvailableTests(result.value);
+        if (result.value.length > 0) {
+          setSelectedPath(result.value[0].path);
+        }
+      } else {
+        console.error(result.error);
+      }
+    } catch (error) {
+      console.error(error);
+    }
```
39bytes, line 57: is this try/catch needed? its kind of ugly to mix result and try/catch

src/renderer/hooks/useTestImport.ts
```diff
@@ -125,3 +125,53 @@ export const useTestImport = () => {
...
+      return err(Error("No tests were found in the specified file."));
+    }
+    return ok(newElems);
+  }
+
+  // Return a function that takes the file path as an argument
+  return async (filePath: string) => {
+    const result = await getTests(filePath);
+    return result;
+  };
```
39bytes, line 176: this is redundant, just return `getTests`

src/renderer/routes/test_sequencer_panel/components/modals/ChangeLinkedTest.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const handleFilePicker = async () => {
+    try {
+      const filePath = await openFilePicker();
+      if (filePath) {
+        await handleDiscoverPytestElements(filePath);
+      }
+    } catch (error) {
+      console.error(error);
+    }
```
39bytes, line 80: then here there's no need to try/catch because `openFilePicker` won't throw

src/renderer/routes/test_sequencer_panel/components/modals/ChangeLinkedTest.tsx
```diff
@@ -0,0 +1,168 @@
...
+      }
+      handleSubmit(selectedPath, testType);
+    } else {
+      window.api.openTestPicker().then((result) => {
+        if (!result) {
+          return;
+        }
+        const { filePath } = result;
+        handleSubmit(filePath, testType);
+      });
```
39bytes, line 96: why not use `openFilePicker` and async/await?

src/renderer/routes/test_sequencer_panel/components/modals/ChangeLinkedTest.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const openFilePicker = async () => {
+    return new Promise<string | null>((resolve, reject) => {
+      window.api.openTestPicker().then((result) => {
+        if (!result) {
+          reject(new Error("No file selected."));
+        } else {
+          resolve(result.filePath);
+        }
+      });
+    });
```
39bytes, line 69: I would just do:
```typescript
const res = await window.api.openTestPicker();
if (!res) return err(Error("No file selected."));
return res.filePath;
```

39bytes (approved):

src/renderer/hooks/useTestSequencerState.ts
LatentDream: It's actually suppose to be `exportToCloud === undefined ? true : exportToCloud,`, I refactored it too fast

39bytes (dismissed):

39bytes (approved):

PR #1169, opened 2024-04-20
Sequencer Gallery

LatentDream:
- [x] Modal
- [x] Load from the Examples folder
- [x] CI Test

# Warning, importing stuff from the Gallery only work in the build. (**not in dev**)


src/renderer/data/apps/sequencer/conditional/Conditional Demo.tjoy
itsjoeoui, line 1: could you remove the while space in file name just in case some bad handling happens haha, sometimes you never know what's gonna happen on windows sometimes

src/renderer/data/apps/sequencer/expected_exported_values/Export & Expected Demo.tjoy
itsjoeoui, line 1: same here :)

src/renderer/routes/test_sequencer_panel/components/modals/SequencerGalleryModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+  setIsGalleryOpen,
+}: AppGalleryModalProps) => {
+  const setOpen = () => {
+    setIsGalleryOpen(true);
+  };
+
+  const useImport = useImportAllSequencesInFolder();
+
+  const handleSequenceLoad = (BaseFolderName: string) => {
+    const relativePath = `src/renderer/data/apps/sequencer/${BaseFolderName}/`;
```
itsjoeoui, line 30: I don't think this is going to work when we package the app tho, this directory certainly will not exist

itsjoeoui (dismissed):

src/renderer/routes/test_sequencer_panel/components/modals/SequencerGalleryModal.tsx
LatentDream: Oh, facts I forgot to test it by packaging the app, it's not the only place in the App using relative path to open the examples. But the packaging might by done with `renderer` only

LatentDream: Oh never mind, the other place just use the raw text

LatentDream: Switched to the same flow as the Test profile. (I'll need to add a "download" fall back if git is not install this week)

LatentDream: Revert to local

src/renderer/routes/test_sequencer_panel/components/modals/SequencerGalleryModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+  DialogContent,
+  DialogHeader,
+  DialogTitle,
+} from "@/renderer/components/ui/dialog";
+import { ScrollArea } from "@/renderer/components/ui/scroll-area";
+import { Separator } from "@/renderer/components/ui/separator";
+import { Button } from "@/renderer/components/ui/button";
+import { useImportAllSequencesInFolder } from "@/renderer/hooks/useTestSequencerProject";
+
+type AppGalleryModalProps = {
```
39bytes, line 12: probably typo in this type name (`SequencerGalleryModalProps`)

src/renderer/routes/test_sequencer_panel/components/modals/SequencerGalleryModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    <Dialog open={isGalleryOpen} onOpenChange={setIsGalleryOpen}>
+      <DialogContent className="flex h-4/5 max-w-5xl flex-col">
+        <DialogHeader>
+          <DialogTitle>
+            <div className="text-3xl">Sequence Gallery</div>
+          </DialogTitle>
+        </DialogHeader>
+        <ScrollArea className="">
+          {data.map((SeqExample) => (
```
39bytes, line 52: should be `camelCase`

src/renderer/routes/test_sequencer_panel/utils/SequenceHandler.ts
```diff
@@ -229,6 +229,9 @@ async function saveToDisk(
 }
 
 async function installDeps(sequence: TestSequencerProject): Promise<boolean> {
+  if (sequence.interpreter.requirementsPath !== null) {
+    return true;
+  }
```
39bytes, line 234: shouldn't this be the other way around? i.e `=== null`

LatentDream: Oh nice catch

39bytes (dismissed):

